PR #32, opened 2020-06-08
Move qeb hwt

pacospace:

- [ ] Yes
- [x] No

Put templates in correct place


pacospace:
> lgtm,
> the webhook-callback-url, would need to be adjusted later with the new cluster url.

yes agree, thanks! :rocket: 

Adding a TODO?

PR #50, opened 2020-06-25
reduce parallelism to 6

KPostOffice:
I tried changing this value to 6 manually but it didn't seem to change the number being run so I'm trying here instead.

### DIFF
--- a/core/overlays/test/configmaps.yaml
+++ b/core/overlays/test/configmaps.yaml
@@ -50,7 +50,7 @@ data:
   config: |
     containerRuntimeExecutor: "k8sapi"
     namespace: "thoth-test-core"
-    parallelism: 7
+    parallelism: 6
 
     # metricsConfig controls the path and port for prometheus metrics
     metricsConfig:


goern:
> I tried changing this value to 6 manually but it didn't seem to change the number being run so I'm trying here instead.
> 
> 

I think you need to restart the workflow controller pod after changing the config map

pacospace:
> > I tried changing this value to 6 manually but it didn't seem to change the number being run so I'm trying here instead.
> 
> I think you need to restart the workflow controller pod after changing the config map

it should be updated automatically actually :) the controller checks for changes in that configmap periodically and put the state in sync :D

goern:
The pre-commit Check needs to pass...

KPostOffice:
@goern 
http://tekton-dashboard-tekton-pipelines.apps.thoth01.lab.upshift.rdu2.redhat.com/#/namespaces/thoth-ci/pipelineruns/aicoe-pipelinerun-5jv9k

harshad16:
> @goern
> http://tekton-dashboard-tekton-pipelines.apps.thoth01.lab.upshift.rdu2.redhat.com/#/namespaces/thoth-ci/pipelineruns/aicoe-pipelinerun-5jv9k

seems to be a flask failure in API request

Issue #52, opened 2020-06-26
ArgoCD App: security-indicators


goern:
As a ThothOps,
I want to have ArgoCD configured to deploy SI workflow components
so that other components like init-job can use them

PR #53, opened 2020-06-26
initial commit of SI app

goern:

implements https://github.com/thoth-station/thoth-application/issues/52

- [ ] Yes
- [x] No


pacospace:
These templates are not correct, https://github.com/thoth-station/si-bandit/pull/8 this seems to be updated for si-bandit, I don't see the PR in si-cloc for the updated one and not sure where the overall workflow template is placed.

pacospace:
> These templates are not correct, [thoth-station/si-bandit#8](https://github.com/thoth-station/si-bandit/pull/8) this seems to be updated for si-bandit, I don't see the PR in si-cloc for the updated one and not sure where the overall workflow template is placed.

The most updated templates are in the namespace already in thoth-test-core on ocp.

security-indicators/base/imagestream.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
pacospace, line 1: we miss imagestreams 

```
- workflow-helpers -> for download-package workflow task
- si-aggreagator -> for aggregator workflow task
```

security-indicators/base/argo-workflows/cloc-template.yaml
```diff
@@ -0,0 +1,67 @@
...
+metadata:
+  name: cloc
+  annotations:
+    thoth-station.ninja/template-version: 0.6.5
+  labels:
+    app: thoth
+    component: si-cloc
+spec:
+  templates:
+    - name: cloc
```
pacospace, line 13: `cloc-from-dir`

security-indicators/base/argo-workflows/bandit-template.yaml
```diff
@@ -0,0 +1,67 @@
...
+metadata:
+  name: bandit
+  annotations:
+    thoth-station.ninja/template-version: 0.1.1
+  labels:
+    app: thoth
+    component: si-bandit
+spec:
+  templates:
+    - name: bandit
```
pacospace, line 13: bandit-from-dir

security-indicators/base/argo-workflows/bandit-template.yaml
```diff
@@ -0,0 +1,67 @@
+---
```
pacospace, line 1: this template is not the correct one

security-indicators/base/security-indicators.yaml
```diff
@@ -0,0 +1,193 @@
+---
```
pacospace, line 1: this should be in openshift folder

goern: do we do that in other apps?

pacospace: one step back then, why is it an app? I thought this is a workflow triggered by another component with logic to do that. Currently, it is not an app right?

for workflow component we use openshift and argo workflows folders in general.

goern: ja, right, which would be the other component that wants to create an SI workflow?

pacospace: <component>-> learn-packages workflow -> solver workflow + SI workflow once graph-sync-solver is completed and checking if that package is already stored on the database.

with empty thoth:
- <component> == init-job initially after provision of thoth uses learn-packages workflow
- <component> == graph-refresh-job uses learn-packages workflow

with thoth with data:
- <component> == graph-refresh-job uses SI workflow for packages solved with no SI evaluate

New tasks:
- [ ] graph schema to sync SI.
- [ ] sync logic in storages for SI.
- [ ] add graph-sync-security-indicator task to SI.
- [ ] create methods to query for SI for a certain package version index.
- [ ] New learn-packages workflow make by solver workflow + si workflow scheduled if some conditions are met.
- [ ] New logic in graph-refresh-job for learn-packages workflow for new knowledge.
- [ ] New logic in graph-refresh-job for si workflow for knowledge already collected.

fridex: I think there does not need to be any logic in init-job. The graph refresh should handle the case when the package is solved but has no si. Otherwise the same logic will be at two different places, but the first one is basically a subset of the latter one.

pacospace: But initi-job logic won't be completely changed, just the workflow to be scheduled, instead of solvers, we have solvers + SI after provision and we can run integrations tests with advises also from the security point of view?

fridex: If you mean the one workflow per each package scheduled, that sounds good.

pacospace: Yes exactly :)

goern:
@pacospace @KPostOffice could one of you please verify that SI is working as expected... in test-thoth-core on ocp cluster. thx!

security-indicators/base/argo-workflows/download-package-template.yaml
```diff
@@ -0,0 +1,39 @@
...
+            value: download_package
+          - name: APP_SCRIPT
+            value: app.sh
+          - name: THOTH_PACKAGE_NAME
+            value: "{{inputs.parameters.THOTH_PACKAGE_NAME}}"
+          - name: THOTH_PACKAGE_VERSION
+            value: "{{inputs.parameters.THOTH_PACKAGE_VERSION}}"
+          - name: THOTH_PACKAGE_INDEX
+            value: "{{inputs.parameters.THOTH_PACKAGE_INDEX}}"
+        image: "workflow-helpers:test"
```
KPostOffice, line 22: This should probably be latest or something

pacospace: `image: workflow-helpers` only actually as we use the local image from the namespace

security-indicators/overlays/test/imagestreamtag.yaml
```diff
@@ -0,0 +1,46 @@
...
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: si-cloc
+spec:
+  tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: quay.io/thoth-station/si-cloc:v0.7.0
```
pacospace, line 42: We miss release on quay of the three images, to put the correct tag.

KPostOffice:
Please add `resubmitPendingPods: true` to all tasks

PR #55, opened 2020-06-30
Add provenance-checker OpenShift template and Argo workflow

fridex:


provenance-checker/base/openshift-templates/provenanceChecker-template.yaml
```diff
@@ -0,0 +1,143 @@
...
+                arguments:
+                  artifacts:
+                    - name: outputdocument
+                      from: "{{tasks.provenance-check.outputs.artifacts.outputdocument}}"
+                  parameters:
+                    - name: "THOTH_DOCUMENT_ID"
+                      value: "{{workflow.parameters.THOTH_DOCUMENT_ID}}"
+                    - name: "THOTH_FORCE_SYNC"
+                      value: "{{workflow.parameters.THOTH_FORCE_SYNC}}"
+
```
harshad16, line 143: the extra newline is causing the pre-commit failure.

provenance-checker/base/argo-workflows/provenance-check-template.yaml
```diff
@@ -0,0 +1,142 @@
...
+              insecure: true
+              accessKeySecret:
+                name: argo-artifact-repository-secrets
+                key: accessKey
+              secretKeySecret:
+                name: argo-artifact-repository-secrets
+                key: secretKey
+      container:
+        name: provenance-check
+        image: quay.io/thoth-station/adviser:v0.6.0-dev
```
pacospace, line 48: this should be simply `adviser` as we use images created in the namespace where the workflow runs.

provenance-checker/base/openshift-templates/provenanceChecker-template.yaml
fridex: Should be fixed, thanks for pointing this out! :+1:

provenance-checker/base/argo-workflows/provenance-check-template.yaml
fridex: Oh, forgot these from testing. Nice catch...

pacospace:
What about the other files required by Argo CD? as in https://github.com/thoth-station/thoth-application/tree/master/adviser for example?

fridex:
> What about the other files required by Argo CD? as in https://github.com/thoth-station/thoth-application/tree/master/adviser for example?

Yes, they will be required. I've added templates I had on disc to have them tracked.

PR #59, opened 2020-06-30
support package-extract project with argocd

harshad16:
support package-extract project with argocd

https://github.com/thoth-station/package-extract/pull/406

- [ ] Yes
- [x] No

includes package-extract application configuration to be deployed via argocd.

## Description

argocd project is to be created for package-extract once the pull request is merged.

### DIFF
--- a/package-extract/overlays/stage/imagestreamtag.yaml
+++ b/package-extract/overlays/stage/imagestreamtag.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: package-extract
+spec:
+  tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: "quay.io/thoth-station/package-extract:v0.6.0-dev"
+      importPolicy: {}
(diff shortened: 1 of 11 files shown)


package-extract/overlays/stage/imagestreamtag.yaml
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: package-extract
+spec:
+  tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: "quay.io/thoth-station/package-extract:v0.6.0-dev"
```
fridex, line 11: How will we manage versions? Will we run sed on all the templates to replace the version string?

goern: for now, its a manual and PR-based process to promote version from test to stage. in test the cyborgs will update as we release to quay.

fridex: ack, thanks for the explanation

PR #71, opened 2020-07-07
Prow roles and rolebinding needed CA perms

harshad16:
Prow roles and rolebinding needed CA perms


goern:
Shall we put this in one overlay that is to be applied with cluster-admin privileged?

harshad16:
> Shall we put this in one overlay that is to be applied with cluster-admin privileged?

great suggestion, will move it around.

harshad16:
@goern can you have a look at this

PR #83, opened 2020-07-08
Decrease beam width for adviser

fridex:

- [ ] Yes
- [x] No

It might be still too large, let's observe any changes.

See https://thoth-station.ninja/docs/developers/adviser/deployment.html#beam-and-it-s-width


goern:
Is it worth setting up some metrics to observe this in a better way?

fridex:
> Is it worth setting up some metrics to observe this in a better way?

This was to reduce OOM. Ideally, we should see more successful responses for larger stacks.

PR #85, opened 2020-07-08
Workflow revsolver

fridex:

- [ ] Yes
- [x] No


revsolver/base/openshift-templates/revsolve_and_sync-template.yaml
```diff
@@ -0,0 +1,116 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: revsolver
+  annotations:
+    description: "Thoth: Reverse Solver  Template"
+    openshift.io/display-name: "Thoth: Reverse Solver"
+    version: 0.6.0
```
harshad16, line 8: let us remove the version so that argo would maintain it.

revsolver/base/openshift-templates/revsolve_and_sync-template.yaml
```diff
@@ -0,0 +1,116 @@
...
+    displayName: "Log reverse solver"
+    required: true
+    value: INFO
+
+objects:
+  - apiVersion: argoproj.io/v1alpha1
+    kind: Workflow
+    metadata:
+      name: "${THOTH_REVSOLVER_WORKFLOW_ID}"
+      annotations:
```
harshad16, line 43: let us remove the version so that argo would maintain it.

revsolver/base/openshift-templates/revsolve_and_sync-template.yaml
fridex: do'h always forgot it. Thanks for spotting this :+1:

PR #93, opened 2020-07-10
Thoth investigator (Consumer) Application

pacospace:

Related-To: https://github.com/thoth-station/investigator/pull/61

- [x] Yes
- [ ] No

Update components name.
Move investigator image to core as it will be used by different component.


investigator/overlays/test/imagestreamtag.yaml
```diff
@@ -1,13 +1,13 @@
...
 metadata:
-  name: unresolved-package-handler
+  name: investigator
 spec:
   tags:
   - name: latest
     from:
       kind: DockerImage
-      name: quay.io/thoth-station/unresolved-package-handler:v0.1.2
+      name: quay.io/thoth-station/investigator:v0.1.2
```
goern, line 10: we have not released it, would the next release be v0.2.0 ?

pacospace: we can do a minor release yes after https://github.com/thoth-station/investigator/pull/48 is merged

investigator/overlays/test/kustomization.yaml
```diff
@@ -5,8 +5,6 @@ commonLabels:
 resources:
   - ../../base
   - thoth-notification.yaml
-patchesStrategicMerge:
```
harshad16, line 8: we need this, the imagestream in overlay would patch into imagestream in base

core/base/imagestreams.yaml
```diff
@@ -17,6 +17,14 @@ spec:
 ---
 apiVersion: image.openshift.io/v1
 kind: ImageStream
+metadata:
```
harshad16, line 20: why are we adding investigator image to the core?
is it because its being used by adviser workflow and investigator deploymentconfig?

pacospace: exactly and later solver will use it in one of the task and graph-refresh-job as well :)

Issue #100, opened 2020-07-10
amun-inspection build run are failling


harshad16:
**Describe the bug**
while amun-inspection build run execute to capture the dockerfile and specification to artifacts. The method to write them to artifacts is causing issue. It check for the command in dockerfile(which is not needed)

```
+ results=/mnt/build/
+ mkdir -p /mnt/build/
+ kubectl logs inspection-test-11ca6956-1-build
+ cat
+ jq .
+ cat
/argo/staging/script: line 10: INSTALL_CMD: unbound variable
```

**To Reproduce**
1. Go to`http://amun-api-thoth-test-core.apps.ocp.prod.psi.redhat.com/api/v1/ui/#/Inspection/post_inspection`
2. Click on 'try-out'
3. See error in the workflow

**Expected behavior**
successful execution of amun-inspection runs.

**Additional context**
The INSTALL_CMD executes as required in the build
https://github.com/thoth-station/amun-api/blob/0e5e1fcfd590869d4f6ea917056c45aa1496808f/amun/dockerfile.py#L52
however while just copying the dockerfile to resultant files, the bash cmd return error stating INSTALL_CMD is unbound,

PR #101, opened 2020-07-10
Fix storing of dockerfile artifact

harshad16:
Fix storing of dockerfile artifact

Fixes: #100 

- [x] No

The copying content of Dockerfile doesn't require bash error checking.

### DIFF
--- a/amun/base/argo-workflows/inspection-build-with-cpu.yaml
+++ b/amun/base/argo-workflows/inspection-build-with-cpu.yaml
@@ -234,13 +234,16 @@ spec:
           results="/mnt/build/"
           mkdir -p "$results"
           kubectl logs "{{inputs.parameters.inspection-id}}-1-build" > "${results}/log"
-          cat <<EOFEOF  | jq . > "${results}/specification"
+          cat << EOF  | jq . > "${results}/specification"
           {{inputs.parameters.specification}}
-          EOFEOF
-          cat <<EOFEOF > "${results}/Dockerfile"
-          {{inputs.parameters.dockerfile}}
-          EOFEOF
+          EOF
 
+          set +e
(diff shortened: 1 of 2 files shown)


amun/base/argo-workflows/inspection-build-with-cpu.yaml
```diff
@@ -234,13 +234,16 @@ spec:
...
           mkdir -p "$results"
           kubectl logs "{{inputs.parameters.inspection-id}}-1-build" > "${results}/log"
-          cat <<EOFEOF  | jq . > "${results}/specification"
+          cat << EOF  | jq . > "${results}/specification"
           {{inputs.parameters.specification}}
-          EOFEOF
-          cat <<EOFEOF > "${results}/Dockerfile"
-          {{inputs.parameters.dockerfile}}
-          EOFEOF
+          EOF
```
goern, line 239: is this EOF intended?

fridex: I think so, it's corresponding to the `cat << EOF`.

amun/base/argo-workflows/inspection-build-with-cpu.yaml
```diff
@@ -234,13 +234,16 @@ spec:
...
 
+          set +e
```
fridex, line 241: What was the issue here?

fridex: Ah, I see. It's a fix to #100. Very nice finding! :100:

fridex: I wonder why bash does not treat all the input as a string until EOF though.

PR #105, opened 2020-07-13
change from block indicator to normal string

harshad16:
change from block indicator to normal string

The block indicator:
```
  keyFormat: >-
    {{workflow.parameters.ceph_bucket_prefix}}/
    {{workflow.parameters.deployment_name}}/
    inspections/
    {{workflow.name}}
```
was creating spaces at the beginning.
as the string is still comparatively small, changed it to inline 

- [ ] Yes
- [x] No


harshad16:
@goern  merging this for now, please let me know if we should make this change differently.

PR #109, opened 2020-07-15
Don't ignore secret yamls

xtuchyna:
This is useless (and unnecessary pain) because we store the secret(s) inside the thoth-application

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-secrets.yaml
-secrets-*.yaml
 application_default_credentials.json
 .vscode
 .env


harshad16:
we use [sops](https://github.com/mozilla/sops#sops-secrets-operations) for secrets creating with encryption, as a developer is creating the secret, the developer would need to first have the secret in the manifest file to run sops on it, we cant have secrets mistakenly pushed to the repo.
please close the pr.

xtuchyna:
aight got it

xtuchyna:
@harshad16 so shouldn't we have this mentioned in readme then?

harshad16:
@xtuchyna yes, we should. please feel free to open a PR with readme update

PR #119, opened 2020-07-16
Drop use-argo from configmaps

fridex:

- [ ] Yes
- [x] No

We no longer support use_argo configuration option to switch to old legacy workload-operator implementation.

### DIFF
--- a/user-api/base/deploymentconfig.yaml
+++ b/user-api/base/deploymentconfig.yaml
@@ -158,11 +158,6 @@ spec:
                 configMapKeyRef:
                   name: user-api
                   key: api-https
-            - name: THOTH_USE_ARGO
-              valueFrom:
-                configMapKeyRef:
-                  name: user-api
-                  key: use-argo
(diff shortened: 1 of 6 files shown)


user-api/base/deploymentconfig.yaml
```diff
@@ -158,11 +158,6 @@ spec:
                 configMapKeyRef:
                   name: user-api
                   key: api-https
-            - name: THOTH_USE_ARGO
```
goern, line 161: its used by more components than user-api...

goern:
It's also used in a few places all over the org... https://github.com/search?q=org%3Athoth-station+THOTH_USE_ARGO&type=Code

fridex:
> It's also used in a few places all over the org... https://github.com/search?q=org%3Athoth-station+THOTH_USE_ARGO&type=Code

They can be removed. The environment variable is not used in the code anymore so providing it does not make any difference.

PR #121, opened 2020-07-16
add another step to workflow to sync results

KPostOffice:
Add syncing for security indicator results to si workflow

Fixes: https://github.com/thoth-station/graph-sync-job/issues/421

### DIFF
--- a/security-indicators/base/security-indicators.yaml
+++ b/security-indicators/base/security-indicators.yaml
@@ -41,7 +41,7 @@ objects:
           - name: THOTH_SECURITY_INDICATORS_AGGREGATION_FUNCTION
             value: "${THOTH_SECURITY_INDICATORS_AGGREGATION_FUNCTION}"
           - name: THOTH_FORCE_SYNC
-            value: "${THOTH_SECURITY_INDICATORS_FORCE_SYNC}"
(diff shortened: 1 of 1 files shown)


security-indicators/base/security-indicators.yaml
```diff
@@ -40,6 +40,8 @@ objects:
             value: "${THOTH_SECURITY_INDICATORS_PACKAGE_INDEX}"
           - name: THOTH_SECURITY_INDICATORS_AGGREGATION_FUNCTION
             value: "${THOTH_SECURITY_INDICATORS_AGGREGATION_FUNCTION}"
+          - name: THOTH_FORCE_SYNC
```
pacospace, line 43: you also need to set in parameters:
```
  - name: THOTH_FORCE_SYNC
    description: Perform force sync - sync data even if they are reported to be existing
    displayName: Perform force sync
    required: true
    value: "0"
```

pacospace, line 43: you also need to set in parameters:
```
  - name: THOTH_FORCE_SYNC
    description: Perform force sync - sync data even if they are reported to be existing
    displayName: Perform force sync
    required: true
    value: "0"
```

PR #134, opened 2020-07-21
deploy pgweb for postgresql web interface

harshad16:
deploy pgweb for postgresql web interface

### DIFF
--- a/core/overlays/test/pgweb.yaml
+++ b/core/overlays/test/pgweb.yaml
@@ -1,0 +1,105 @@
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: pgweb
+spec:
+  lookupPolicy:
+    local: true
+  tags:
+  - annotations: null
+    from:
+      kind: DockerImage
+      name: sosedoff/pgweb:0.11.3
+    name: latest
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    service: pgweb
+  name: pgweb
+spec:
+  replicas: 1
+  selector:
+    service: pgweb
+  test: false
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: thoth
+        app.kubernetes.io/component: core
+        app.kubernetes.io/managed-by: thothops-argocd
+        service: pgweb
+    spec:
+      containers:
+      - image: 'pgweb:latest'
+        env:
+        - name: DB_HOST
+          value: postgresql
+        - name: DB_PORT
+          value: "5432"
+        - name: DB_DATABASE
+          valueFrom:
+            secretKeyRef:
+              name: postgresql
+              key: database-name
+        - name: DB_USER
+          valueFrom:
+            secretKeyRef:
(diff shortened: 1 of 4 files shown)


core/overlays/test/pgweb.yaml
```diff
@@ -0,0 +1,117 @@
...
+          httpGet:
+            path: '/'
+            port: 8081
+            scheme: HTTP
+          initialDelaySeconds: 10
+          periodSeconds: 60
+          timeoutSeconds: 10
+  test: false
+  triggers:
+  - type: ConfigChange
```
goern, line 81: configChange trigger is a little bit tricky, lets see how it behaves...

core/overlays/test/pgweb.yaml
```diff
@@ -0,0 +1,117 @@
...
+kind: DeploymentConfig
+metadata:
+  labels:
+    service: pgweb
+    app: thoth
+    component: pgweb
+  name: pgweb
+spec:
+  replicas: 1
+  selector:
```
goern, line 25: https://github.com/thoth-station/thoth-application/blob/95d6dea7e6015c5e37dd9740c0f72695f64d5a4f/core/overlays/test/kustomization.yaml#L3 will add more labels, so the selector and the template labels need to include them. feel free to remove the version commonLabel ;)

core/overlays/test/pgweb.yaml
```diff
@@ -0,0 +1,117 @@
...
+        - pgweb
+      from:
+        kind: ImageStreamTag
+        name: 'pgweb:latest'
+---
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: pgweb
+  labels:
```
goern, line 95: we dont need these

core/overlays/test/pgweb.yaml
```diff
@@ -0,0 +1,117 @@
...
+    component: pgweb
+spec:
+  to:
+    kind: Service
+    name: pgweb
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
```
goern, line 106: see other comments on commonLabel

core/overlays/test/pgweb.yaml
harshad16: i have removed this for now, as we should first fix this in amun-api or user-api, then we can replicate the behavior in other deployment configs.

core/overlays/test/pgweb.yaml
harshad16: Added the commonlabel to service's selector and template's label.

goern:
Do we really need this in stage, or can we life w/o it?

@fridex @pacospace

fridex:
> Do we really need this in stage, or can we life w/o it?
> 
> @fridex @pacospace

I'm fine without having it. We have periodic dumps that can be downloaded, no need to expose stage database this way - users should not have it accessible.

harshad16:
we do want it for the test though right?

fridex:
> we do want it for the test though right?

Yes, let's keep it for the test environment.

Issue #139, opened 2020-07-22
mi-scheduler SA needs permission


goern:
**Describe the bug**
`HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"templates.template.openshift.io is forbidden: User \\"system:serviceaccount:thoth-test-core:mi-scheduler\\" cannot list resource \\"templates\\" in API group \\"template.openshift.io\\" in the namespace \\"thoth-test-core\\"","reason":"Forbidden","details":{"group":"template.openshift.io","kind":"templates"},"code":403}\n'`

PR #140, opened 2020-07-22
updated MI scheduler role

goern:

fixes https://github.com/thoth-station/thoth-application/issues/139

- [ ] Yes
- [X] No


mi-scheduler/base/role.yaml
```diff
@@ -3,10 +3,347 @@ kind: Role
...
+      - templates
+    verbs:
+      - create
+      - delete
+      - deletecollection
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
harshad16, line 214: Feels like mi-scheduler only require these permissions. 
we should keep a close bound on roles based on requirements. 
WDYT?

harshad16:
oh, the permission to access configmaps are needed by mi-scheduler, I missed that and didn't realize this pr has them removed.
will open a commit to fix that.

PR #151, opened 2020-07-24
add image stream and WorkflowTemplate for thoth-messaging

KPostOffice:
We need to be able to use the messaging `CLI` as a step in workflows.


harshad16:
ah, which component would be triggering messaging ?

messaging/base/argo-workflows/send-message.yaml
```diff
@@ -0,0 +1,51 @@
+---
```
pacospace, line 1: I think this can go to core actually as many workflows can use it? @goern @harshad16

harshad16: yeah, I agree, if this is not a standalone component. if this triggered by some other component then this should go in that application subdirectory.

pacospace: It can be used as workflow-helpers in different components basically :)

KPostOffice: Should I also add a sub-directory `argo-workflows` in `core/base`? To put the WorkflowTemplate in?

harshad16: yeah that would be appropriate space to add these

pacospace:
> ah, which component would be triggering messaging ?

Solver will use it and Adviser also when we will migrate the current task

core/base/argo-workflows/send-message.yaml
```diff
@@ -0,0 +1,51 @@
...
+            value: "{{inputs.parameters.THOTH_MESSAGING_CREATE_IF_NOT_EXISTS}}"
+          - name: THOTH_MESSAGING_MESSAGE_CONTENTS
+            value: "{{inputs.parameters.THOTH_MESSAGING_MESSAGE_CONTENTS}}"
+          - name: THOTH_MESSAGING_PARTITIONS
+            value: "{{inputs.parameters.THOTH_MESSAGING_PARTITIONS}}"
+          - name: KAFKA_BOOTSTRAP_SERVERS
+            valueFrom:
+              configMapKeyRef:
+                name: KAFKA_BOOTSTRAP_SERVERS
+                key: kafka-bootstrap-servers
```
pacospace, line 36: `kafka`

pacospace:
```
          - name: KAFKA_BOOTSTRAP_SERVERS
            valueFrom:
              configMapKeyRef:
                key: kafka-bootstrap-servers
                name: kafka
```

PR #171, opened 2020-07-29
Amun API URL used in dependency monkey needs scheme

fridex:
Otherwise, Dependency Monkey stores result in a file called `amun-api.thoth-test-core.svc` as it fails to detect service URL.

### DIFF
--- a/amun/overlays/test/configmaps.yaml
+++ b/amun/overlays/test/configmaps.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 data:
-  amun-api-url: amun-api.thoth-test-core.svc
+  amun-api-url: http://amun-api.thoth-test-core.svc
   deployment-name: ocp-prod-test
   infra-namespace: thoth-test-core
   inspection-namespace: thoth-test-core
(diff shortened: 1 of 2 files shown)


amun/overlays/test/configmaps.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 data:
-  amun-api-url: amun-api.thoth-test-core.svc
+  amun-api-url: http://amun-api.thoth-test-core.svc
```
harshad16, line 4: does adding the prefix `http://` to the service , would solve the issue?
or should we use the actual route like `https://amun.test.thoths-station.ninja`?

fridex: It looks like http works within the cluster (tested from Amun API pod):

```
(app-root) sh-4.4$ curl https://amun-api.thoth-test-core.svc
curl: (7) Failed to connect to amun-api.thoth-test-core.svc port 443: No route to host
```

```
(app-root) sh-4.4$ curl http://amun-api.thoth-test-core.svc
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to target URL: <a href="api/v1/ui">api/v1/ui</a>.  If not click the link.
```

harshad16: thanks for the insight

PR #172, opened 2020-07-29
place adviser templates in infra namespace

harshad16:
place adviser templates in infra namespace

### DIFF
--- a/adviser/overlays/stage/imagestreamtag.yaml
+++ b/adviser/overlays/stage/imagestreamtag.yaml
@@ -8,7 +8,7 @@ spec:
     - name: latest
       from:
         kind: DockerImage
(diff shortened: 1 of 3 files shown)


harshad16:
@fridex dependency-monkey uses adviser image and management-api tries to trigger it in middletier namespace(also in arch.dia).
should the dependency-monkey workflow be available in both middletier and backend?

fridex:
> @fridex dependency-monkey uses adviser image and management-api tries to trigger it in middletier namespace(also in arch.dia).
> should the dependency-monkey workflow be available in both middletier and backend?

Oh, so the original design was to have dependency-monkey in the middletier namespace and adviser in the backend namespace. With a local lookup policy adviser image would need to be present in both, middletier and backend namespace.

PR #176, opened 2020-07-29
Add view verb

xtuchyna:
## Description
mi-scheduler cronjob currently still fails. A part of the error is:
 ```User \\"system:serviceaccount:thoth-test-core:mi-scheduler\\" cannot list resource \\"templates\\" in API group \\"template.openshift.io\\" in the namespace \\"thoth-test-core\\"","reason":"Forbidden","details":{"group":"template.openshift.io","kind":"templates"},"code":403}\n'```
and this could mean that the role currently set is not enough.

mi-scheduler has `list` added to the verbs and also apigroup `template.openshift.io` specified (https://github.com/thoth-station/thoth-application/blob/6818ef5fee948f46d56c116ad6e9a0e3bdd4bff6/mi-scheduler/base/role.yaml#L11-L17), however the openshift says otherwise.

was texting w/ @fridex and he suggested adding `view`, so maybe that could help

## Details
```2020-07-29 11:04:13,771   1 CRITICAL root:102: Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/openshift/dynamic/client.py", line 42, in inner
    resp = func(self, *args, **kwargs)
  File "/opt/app-root/lib/python3.6/site-packages/openshift/dynamic/client.py", line 245, in request
    _return_http_data_only=params.get('_return_http_data_only', True)
  File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 355, in request
    headers=headers)
  File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/rest.py", line 231, in GET
    query_params=query_params)
  File "/opt/app-root/lib/python3.6/site-packages/kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (403)
Reason: Forbidden
[...]

PR #179, opened 2020-07-29
Change templates' specifications, remove view verb & add workflow ID

xtuchyna:


goern:
why could this be removed?

/hold

xtuchyna:
>why could this be removed?
It was added based on this PR #176 but we have found out that it was not the right solution.

harshad16:
> why could this be removed?
> 
> /hold

i checked it, verb view is not present for template, also get would do the needful.

mi-scheduler/base/openshift-templates/mi-run-template.yaml
```diff
@@ -3,19 +3,26 @@ kind: Template
...
+
 objects:
   - apiVersion: argoproj.io/v1alpha1
     kind: Workflow
     metadata:
-      generateName: mi-analysis-
+      name: mi-analysis-${WORKFLOW_ID}
       labels:
         app: scropsmetrics
     spec:
```
harshad16, line 28: Include arguments in spec as well.

```
arguments:
   parameters:
   - name: REPOSITORY
     value: ${REPOSITORY}
```

harshad16:
Also can you add the below provided rolebinding in https://github.com/thoth-station/thoth-application/blob/142cc216331e4afb2dadcdcabba6704b7b10c34b/mi-scheduler/base/role_binding.yaml

```
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: mi-scheduler-argo-admin
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: argo-admin
subjects:
  - kind: ServiceAccount
    name: mi-scheduler
```

mi-scheduler/base/role.yaml
```diff
@@ -27,4 +27,4 @@ rules:
       - patch
       - update
       - watch
-      - view
+
```
harshad16, line 30: remove the extra newline

PR #182, opened 2020-07-29
upgrade postgresql pvc

harshad16:
upgrade postgresql pvc
from 32Gi to 64Gi, we have 120 Gi storage limit in the clusters.

### DIFF
--- a/core/overlays/graph-stage/postgresql.yaml
+++ b/core/overlays/graph-stage/postgresql.yaml
@@ -28,7 +28,7 @@ items:
         - ReadWriteOnce
       resources:
         requests:
-          storage: 32Gi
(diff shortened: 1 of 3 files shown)


core/overlays/graph-stage/postgresql.yaml
```diff
@@ -28,7 +28,7 @@ items:
         - ReadWriteOnce
       resources:
         requests:
-          storage: 32Gi
+          storage: 64Gi
```
fridex, line 31: Can we do even more considering the current quota? E.g. `64 + 32 = 96Gi`?

harshad16: yes we came

PR #184, opened 2020-07-29
Buildconfig for graph-sync for developement purpose

harshad16:
Buildconfig for graph-sync for developement purpose

### DIFF
--- a/graph-sync/overlays/test/buildconfig.yaml
+++ b/graph-sync/overlays/test/buildconfig.yaml
@@ -1,0 +1,50 @@
+---
+apiVersion: v1
+kind: BuildConfig
+metadata:
+  labels:
+    app: thoth
+    component: graph-sync
+  name: graph-sync-job
+spec:
+  resources:
+    limits:
+      cpu: 2000m
+      memory: 768Mi
+    requests:
+      cpu: 2000m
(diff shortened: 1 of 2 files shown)


graph-sync/overlays/test/buildconfig.yaml
```diff
@@ -0,0 +1,50 @@
...
+    limits:
+      cpu: 2000m
+      memory: 768Mi
+    requests:
+      cpu: 2000m
+      memory: 768Mi
+  output:
+    to:
+      kind: ImageStreamTag
+      name: "graph-sync-job:latest"
```
KPostOffice, line 20: should we push to test?

harshad16: the argocd application wouldnt allow us to use the test tag in graph-sycn workflow , it would overide it to latest.
may be for now it is okay to use latest.

@goern what are your thoughts on this?

goern: hmm, or we reconfig the argocd to not self-heal? so that users can change the workflow temporarily?

goern: erm, ja :) one or the other... idk :)

PR #191, opened 2020-07-30
amun workflow requires pods patch access

harshad16:
amun workflow requires pods patch access

Forgot to add this in pervious PR: #190


harshad16:
merging this for testing

PR #193, opened 2020-07-30
Resubmit pending pods in Amun inspections

fridex:

- [x] No

### DIFF
--- a/amun/base/openshift-templates/inspection-with-cpu-workflow.yaml
+++ b/amun/base/openshift-templates/inspection-with-cpu-workflow.yaml
@@ -101,6 +101,7 @@ objects:
           templateRef:
             name: inspection-build-with-cpu-template
             template: main
+          resubmitPendingPods: true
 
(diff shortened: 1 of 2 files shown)


fridex:
@KPostOffice we use `steps` and not dag here, not sure if the configuration option applies here as well.

KPostOffice:
> @KPostOffice we use `steps` and not dag here, not sure if the configuration option applies here as well.

It does. I still don't think it was implemented properly yet, I can link the discussion/PR from Argo if you want

fridex:
> > @KPostOffice we use `steps` and not dag here, not sure if the configuration option applies here as well.
> 
> It does. I still don't think it was implemented properly yet, I can link the discussion/PR from Argo if you want

It does not look like this configuration is propagated correctly when applying this configuration to the cluster. When I adjust the template, this configuration option is not shown in the workflows created (si correctly states it in the workflow though).

PR #197, opened 2020-07-30
Provide workflow parallelism in workflow defaults section

fridex:

- [x] No


fridex:
We might want to configure default TTL here as well to keep namespaces clean.

PR #207, opened 2020-08-03
Decrease solver TTL

fridex:

Let's decrease solver TTL so that solvers finish faster and we can check number of solvers pending.


fridex:
@harshad16 PTAL

Issue #187, opened 2020-07-30
release 0.7.0 blocker


goern:
- [ ] new patch release common
- [ ] new patch release storages
- [ ] new patch release solver
- [ ] new patch release graph-sync
- [ ] new patch release init-job
- [ ] validate thoth-application config for stage to use the latest releases
- [ ] validate ceph bucket configuration
- [ ] clear ceph bucket
- [ ] validate deployment to ocp-stage
- [ ] delete postgresql PV
- [ ] kick off init-job

PR #216, opened 2020-08-06
setup backend namespace with argo workflow controller

harshad16:
setup backend namespace with argo workflow controller

Related-To: #187 

- [x] No

- include workflow-controller manifest in backend-namespace to setup argo
- update configmaps and kustomize.yaml

## Description

- setup backend namespace for adviser workflow to execute.
- allow all the data to be stored in ocp-{stage|test} ceph bucket.

### DIFF
--- a/core/overlays/stage/configmaps.yaml
+++ b/core/overlays/stage/configmaps.yaml
@@ -6,10 +6,10 @@ metadata:
 data:
   applicationVersion: 0.7.0
   deployment-name: ocp-stage
-  amun-api-url: "UNSET"
+  amun-api-url: "http://amun-api-thoth-amun-api-stage.apps.ocp.prod.psi.redhat.com"
   amun-inspection-namespace: thoth-amun-inspection-stage
   amun-namespace: thoth-amun-api-stage
-  backend-namespace: aicoe-infra-prod
+  backend-namespace: thoth-backend-stage
   middletier-namespace: thoth-middletier-stage
   frontend-namespace: thoth-frontend-stage
   infra-namespace: thoth-infra-stage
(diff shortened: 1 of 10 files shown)


core/overlays/stage/configmaps.yaml
```diff
@@ -23,7 +23,7 @@ data:
   postgresql-host: pgbouncer.thoth-graph-stage.svc
   rsyslog-host: ""
   rsyslog-port: ""
-  storage-bucket-name: ocp-prod-stage
+  storage-bucket-name: ocp-stage
```
pacospace, line 26: why do we not use `ocp-prod-stage` here?

core/overlays/middletier-stage/configmaps.yaml
```diff
@@ -25,8 +26,8 @@ data:
       archiveLogs: False
 
       s3:
-        bucket: "DH-DEV-DATA"
-        keyPrefix: "data/thoth/thoth-middletier-stage/argo/artifacts/"
+        bucket: "thoth"
+        keyPrefix: "data/ocp-stage/argo/artifacts/"
```
pacospace, line 30: `ocp-prod-stage`?

core/overlays/stage/configmaps.yaml
harshad16: either one is fine for me, we should just stick with one.

core/overlays/middletier-stage/configmaps.yaml
harshad16: if we keep ocp-stage, it would have same consistency as 
<cluster-name>-<env-type> 
Ex: ocp-stage, ocp-test, moc-prod, psi-stage, psi-test 
earlier we had prefix of thoth , like thoth-psi-test , we can have it that way too

pacospace: Sounds good to me :) Thanks!

PR #225, opened 2020-08-07
updated manifest files based on requirement

harshad16:
updated manifest files based on the requirement


adviser/base/openshift-templates/dependency-monkey.yaml
```diff
@@ -45,43 +42,34 @@ parameters:
     displayName: Beam width
     value: "1000000"
   - name: THOTH_DEPENDENCY_MONKEY_STACK_OUTPUT
-    required: false
```
pacospace, line 48: why do we remove them?

harshad16: argocd fails to deploy as it sees this variable as default if set to false. 
showed this issue during the demo as well.

pacospace: You explained it today very well today indeed :) Thanks!! Yesterday I was still ignorant :D

PR #257, opened 2020-08-17
stage deployment required manifest files

harshad16:

- [x] No

- manifest files for package-update stage deployment 

## Description

stage deployment required manifest files


package-update/overlays/stage/imagestreamtag.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+    app: thoth
+    component: package-update
+  name: package-update-job
+spec:
+  tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: quay.io/thoth-station/package-update-job:v0.6.0-dev
```
goern, line 13: should we start with v0.7.0?

package-update/overlays/stage/imagestreamtag.yaml
```diff
@@ -0,0 +1,16 @@
...
+      importPolicy: {}
+      referencePolicy:
+        type: Source
```
goern, line 16: a) is this still a valid option on ocp4.3 ? and b) should we reconfigure stage to use `local` to reduce coupling with quay?

package-update/overlays/stage/imagestreamtag.yaml
harshad16: thanks for pointing it.

package-update/overlays/stage/imagestreamtag.yaml
harshad16: thanks for pointing it.

PR #277, opened 2020-08-20
Add more CPU to Amun

fridex:

- [x] No

## Description

Amun is slow to start, let's add more CPU.


fridex:
The issue was caused by a key in config map not available, hence the deployment failed. I think we could keep CPU that low for now.

PR #279, opened 2020-08-20
Bump Amun API in stage version to 0.9.1

fridex:

- [x] No

Depends-On: https://github.com/thoth-station/amun-api/issues/541


fridex:
I will merge this one in order to continue with work. Feel free to raise any comments or suggestions.

Issue #298, opened 2020-08-24
Prometheus metrics name do not allow - in name


pacospace:
**Describe the bug**
Prometheus metrics name do not allow `-` in name

**To Reproduce**
1. Schedule SLO-reporter with THOTH_ENVIRONMENT='ocp-stage'

**Expected behavior**
SLO Reporter stores metrics successfully on Ceph

**Screenshots**
```
17 lines
INFO:thoth.slo_reporter:Considering interval for metrics of 1 day/s.
INFO:thoth.slo_reporter:THOTH_EVALUATION_METRICS_NUMBER_DAYS set to 1.
INFO:thoth.slo_reporter:Interval: 2020-08-23 - 2020-08-24
Traceback (most recent call last):
  File "app.py", line 328, in <module>
    main()
  File "app.py", line 323, in main
    day_of_week=day_of_week,
  File "app.py", line 254, in run_slo_reporter
    configuration = Configuration(start_time=start_time, end_time=end_time, number_days=number_days, dry_run=dry_run)
  File "/opt/app-root/src/thoth/slo_reporter/configuration.py", line 75, in __init__
    registry=self.prometheus_registry,
  File "/opt/app-root/lib/python3.6/site-packages/prometheus_client/metrics.py", line 339, in __init__
    labelvalues=labelvalues,
  File "/opt/app-root/lib/python3.6/site-packages/prometheus_client/metrics.py", line 108, in __init__
    raise ValueError('Invalid metric name: ' + self._name)
ValueError: Invalid metric name: thoth_sli_weekly_ocp-test
```

**Additional context**

PR #299, opened 2020-08-24
Add missing env in CM

pacospace:

Fixes: https://github.com/thoth-station/thoth-application/issues/298

- [ ] Yes
- [x] No

Use correct environment variable to avoid Prometheus errors.

### DIFF
--- a/slo-reporter/base/cronjob.yaml
+++ b/slo-reporter/base/cronjob.yaml
@@ -56,7 +56,7 @@ spec:
                   valueFrom:
                     configMapKeyRef:
                       name: thoth
-                      key: deployment-name
+                      key: environment
                 - name: THOTH_MIDDLETIER_NAMESPACE
                   valueFrom:
                     configMapKeyRef:
--- a/core/overlays/stage/configmaps.yaml
+++ b/core/overlays/stage/configmaps.yaml
@@ -6,6 +6,7 @@ metadata:
 data:
   applicationVersion: 0.7.0
   deployment-name: ocp-stage
+  environment: stage
   amun-api-url: "http://amun-api-thoth-amun-api-stage.apps.ocp.prod.psi.redhat.com"
   amun-inspection-namespace: thoth-amun-inspection-stage
   amun-namespace: thoth-amun-api-stage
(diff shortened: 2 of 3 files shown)


slo-reporter/base/cronjob.yaml
```diff
@@ -56,7 +56,7 @@ spec:
                   valueFrom:
                     configMapKeyRef:
                       name: thoth
-                      key: deployment-name
+                      key: environment
```
fridex, line 59: Why not keep using deployment name?

fridex: OK, I see it's related to #298. So what about replacing `-` with `_` or split on `-` following the `deployment_name` naming convention? Introducing another similar variable can cause confusion.

pacospace: Two reasons:

1. `THOTH_ENVIRONMENT`  variable is used to create metric for info.

```
self.environment = os.environ["THOTH_ENVIRONMENT"]
```

```
            self.thoth_weekly_sli = Gauge(
                f"thoth_sli_weekly_{self.environment}",
                "Weekly Thoth Service Level Indicators",
                ["sli_type", "metric_name"],
                registry=self.prometheus_registry,
            )
```
Prometheus metrics do not allow metrics name with `-`, therefore `ocp-stage` cannot be used.

This could be solved by adding `self.environment.replace('-', "_")`

2. but there is a second issue, Prometheus adds label to metrics from DH side. Therefore they contain `job="Thoth Metrics (stage)"` where `test`, `stage`, `prod` is used to differentiate same metric name.
And because `THOTH_ENVIRONMENT`  variable is used to retrieve metrics from Prometheus, even applying solution at 1. `ocp-stage` will not lead to any metrics because they do not exist on Prometheus side

Therefore this would require changes in how metrics are retrieved from DH side where we make sure `job="Thoth Metrics (stage)"` contains the same name as deployment name.

To summarize:

- the current solution in this PR is easy and fast, second solution which requires different changes in different places and new releases can be done as well of course, but it will take some time.

We can create some cards to change that. WDYT? @fridex

fridex: I think having the split method will not introduce tech dept - we already state the environment in `deployment_name` which is used in the whole Thoth application.

pacospace: https://github.com/thoth-station/slo-reporter/pull/70

pacospace: In this PR I will remove environment variable not used anymore

PR #311, opened 2020-08-25
on exit send message

KPostOffice:
https://github.com/thoth-station/amun-api/issues/552

### DIFF
--- a/amun/base/openshift-templates/inspection-with-cpu-workflow.yaml
+++ b/amun/base/openshift-templates/inspection-with-cpu-workflow.yaml
@@ -63,12 +63,20 @@ objects:
         value: "${PHYSICAL_CPUS}"
       - name: processor
         value: "${PROCESSOR}"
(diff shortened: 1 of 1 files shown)


goern:
mazbe the manifests repo is not the right place to store openshift template or workflow templates

goern (approved):

fridex:
I'm probably missing here where the `send-message` template is. Also, could this be applied to the template that is not specific to CPU? I guess we need to add thoth-messaging component into Amun deployment and kafka configuration to make this work.

harshad16:
> I'm probably missing here where the `send-message` template is. Also, could this be applied to the template that is not specific to CPU? I guess we need to add thoth-messaging component into Amun deployment and kafka configuration to make this work.

`send-message` template is in **core** directory, please copy that agro workflowtemplate to amun base and also the messaging imagestreamtag, if we gonna use it in amun.

PR #318, opened 2020-08-26
Increase parallelism to see how resources are handled

fridex:

Related: https://github.com/thoth-station/thoth-application/pull/316

- [x] No

## Description

As self-healing is turned on, let's make this change here to see how argo behaves.


fridex:
Merging this to test the behavior.

PR #336, opened 2020-08-27
Try to setup secondsAfterCompletion to clean up workflows

fridex:

- [x] No

I tried to debug why workflows do not get deleted. It looks like service account has correct perms and all the rest looks configured correctly. Let's try to experiment with `secondsAfterCompletion` and see if it helps.


pacospace (approved):
:+1: sounds and we should not have a problem with SLI/SLO as we will get metrics from Argo workflows about success/fails/ etc..

PR #348, opened 2020-09-01
Bump management-api to v0.7.0 in stage

fridex:
## Related Issues and Dependencies

Related: https://github.com/thoth-station/management-api/issues/560

## This introduces a breaking change

- [x] No

PR #349, opened 2020-09-01
Revert "Bump management-api to v0.7.0 in stage"

fridex:
Reverts thoth-station/thoth-application#348

I will try to revert this, inspections do not run, no workflows are scheduled. Checking if this update could cause it.


fridex:
Yes, it was caused by this update:

```
 Failed: invalid spec: templates.dependency-monkey.tasks.dm templates.dm inputs.parameters.THOTH_ADVISER_PREDICTOR_CONFIG was not supplied
```

PR #355, opened 2020-09-02
Provide secondsAfterCompletion in all workflows

fridex:

- [x] No


fridex:
After debugging, it looks like workflows get properly deleted. The only ones that stay are the ones that failed with `pod deleted`.

pacospace (approved):
/lgtm

PR #358, opened 2020-09-02
Set rebumitPendingPods to all the workflows

fridex:

Fix `pod deleted` issues in workflows.

Related: https://github.com/argoproj/argo/issues/3918 https://github.com/argoproj/argo/issues/1913#issuecomment-685795749

- [x] No


fridex:
It's not clear to me if this should be applied per template or per workflow label (the issue states both are valid?!).

pacospace (approved):
Thanks!!

PR #361, opened 2020-09-03
Move resubmit pending pods to task definitions

fridex:

Related: https://github.com/thoth-station/thoth-application/pull/358

- [x] No


fridex:
It looks like it does not get applied on the workflow level.

pacospace (approved):

PR #362, opened 2020-09-03
Add resubmitPendingPods to all templates in workflow

fridex:

Related: https://github.com/thoth-station/thoth-application/pull/358 https://github.com/thoth-station/thoth-application/pull/361

- [x] No


fridex:
The issue still persists. Let's try applying to config option to templates.

fridex:
It looks like this solution works.

PR #368, opened 2020-09-03
Increase solver CPU resources and time allocated

fridex:

- [ ] Yes
- [x] No


fridex:
I spotted few failures due to not enough time allocated. Let's give it more CPU and time.

pacospace (approved):
:100:

PR #369, opened 2020-09-03
Always perform garbage collection of pods

fridex:

We do not need to keep objects in the cluster, they can be garbage collected. Otherwise, they occupy quota. All the relevant information regarding failures are reported into Sentry.

- [x] No


fridex:
The initializer workflow keeps this configuration option (can be handy on green field deployment), I turned this off for bulk workflows only.

pacospace (approved):
:beer:

PR #373, opened 2020-09-07
Reduce workflow parallelism to 1 in amun-inspection-stage

fridex:
... as amun inspections run builds, Argo has no control on resources used by
builds. This leads to issues when a workflow is stuck waiting for a build that
failed because of resources available in the cluster.

- [x] No


pacospace (approved):
/lgtm

Issue #111, opened 2020-07-16
User API autoscaler


fridex:
**Is your feature request related to a problem? Please describe.**

As reported originally in https://github.com/thoth-station/user-api/issues/739 we could configure autoscaler for user API.

**Describe the solution you'd like**

Configure autoscaler so that user API is automatically scaled up if too many requests are made at the same time. As a start, we can use CPU utilization and observe how the API works. Later, we can utilize metrics, such as the number of connections to scale the replicas up/down.

**Describe alternatives you've considered**

Have constantly scaled multiple user API replicas, but that does not scale automatically based on load.

**Additional context**

https://docs.openshift.com/container-platform/3.9/dev_guide/pod_autoscaling.html
https://github.com/thoth-station/user-api/issues/739

VannTen:
from my experience, web server like load scaling on cpu usage is not super reliable. Scaling on requests rate is probably a better idea, and the metrics is probably already available from whatever python webserver implementation.

harshad16:
**Acceptance criteria**
- [ ] Check for custom metrics adapter on the cluster.  
- [ ] We can use the HTTP metrics from user-api usage and plug that in horizontal auto scaler
- [ ] create a panel in grafana dashboard for user-api instances.

**Resource**
- https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-metrics-apis
- https://docs.openshift.com/container-platform/4.11/nodes/pods/nodes-pods-autoscaling-custom.html

story points: 3pt

PR #376, opened 2020-09-07
WIP: Add horizontal pod autoscaler for user-api

fridex:

Fixes: https://github.com/thoth-station/thoth-application/issues/111

- [x] No

### DIFF
--- a/user-api/base/deploymentconfig.yaml
+++ b/user-api/base/deploymentconfig.yaml
@@ -177,3 +177,16 @@ spec:
             periodSeconds: 60
             timeoutSeconds: 10
   test: false
+---
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: user-api
+spec:
+  maxReplicas: 2
+  minReplicas: 1
+  scaleTargetRef:
+    apiVersion: apps.openshift.io/v1
+    kind: DeploymentConfig
+    name: user-api
+  targetCPUUtilizationPercentage: 75


user-api/base/deploymentconfig.yaml
```diff
@@ -177,3 +177,16 @@ spec:
...
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: user-api
+spec:
+  maxReplicas: 2
+  minReplicas: 1
+  scaleTargetRef:
+    apiVersion: apps.openshift.io/v1
+    kind: DeploymentConfig
```
fridex, line 190: This can be one of the drawbacks of using `Deployment` instead of `DeploymentConfig` for Argo CD. `HorizontalPodAutoscaler` works just for `DeploymentConfig` and `ReplicationController` [1].

[1] https://docs.openshift.com/container-platform/4.5/nodes/pods/nodes-pods-autoscaling.html

goern: as we have a DC, let's test the HPA... @harshad16 wdyt?

fridex: @harshad16 btw, what about using webhook triggers for triggering rollouts of DeploymentConfig from argo cd? Would that work?

goern: maybe @HumairAK has an updated view and some insights on the support of DeploymentConfigs by ArgoCD too?

fridex: If not directly from argocd maybe we could do it from the build pipeline (a specific "step" for dc)?

harshad16: @goern , yes lets test Horizontal pod autoscaler.

@fridex , we can trigger the deploymentconfig through a trigger, would check it, if argocd can do this, as it would something for all. or else we can work having this triggered from pipeline.

goern:
@fridex @harshad16 pinging this again, shall we do that for cnv-prod?

fridex:
> @fridex @harshad16 pinging this again, shall we do that for cnv-prod?

I don't know if that would be required now. Do we want to do tests for the user API to see how many requests we can serve for `replicas=1`?

goern:
what is the status on this one? shall we postpone/close or get it merged this sprint?

/kind demo

goern:
@fridex shall we close this and re-open if https://github.com/thoth-station/thoth-application/issues/1769 is done?

fridex:
> @fridex shall we close this and re-open if #1769 is done?

Yes, I think we can close this. It is documented here and this is not required right now.

/close

PR #378, opened 2020-09-08
Increase parallelism for the current inspection batch

fridex:
... to utilize cluster resources better and speed up the batch

- [x] No


harshad16 (approved):
lgtm

PR #379, opened 2020-09-08
Increase parallelism even more

fridex:
## Related Issues and Dependencies

See https://github.com/thoth-station/thoth-application/pull/378

## This introduces a breaking change

- [x] No

PR #380, opened 2020-09-08
Increase parallelism even more

fridex:

It looks like we can do even more in parallel.

Related: #378 #379

- [x] No


harshad16 (approved):
lgtm

PR #381, opened 2020-09-08
Reduce parallelism in thoth-middletier-stage

fridex:

It looks like workflow-controller goes into some deadlock after some time - pods are not scheduled due to resources but also, the finished ones do not proceed. I suspect it's some bug in resources handling. Let's try to decrease parallelism.

- [x] No


harshad16 (approved):
lgtm
we can have requirement computed based on worklfows and calculate the parallelism based on that, for future updates.

fridex:
> lgtm
> we can have requirement computed based on worklfows and calculate the parallelism based on that, for future updates.

Ideally have this baked in argo, it is getting annoying considering our scheduling units have parameterized resource requests

PR #382, opened 2020-09-08
Try to clean up workflows as soon as possible

fridex:

- [x] No


fridex:
Let's try to reduce TTL so that namespace is cleaned as soon as possible and we respect `count/pod` limtis.

harshad16 (approved):
lgtm

Issue #422, opened 2020-09-14
Modify package-release component to use the new implementation as producer


pacospace:
**Is your feature request related to a problem? Please describe.**
See new package release producer: https://github.com/thoth-station/package-releases-job/issues/446

**Additional context**

pacospace:
is tested successfully?

bissenbay:
> is tested successfully? @bissenbay

@pacospace I'm waiting for this release https://github.com/thoth-station/package-releases-job/issues/484, then I would like to test again cause it was failing with the current version of `thoth-messaging` which is `0.7.3`.

bissenbay:
@pacospace So, I tested by running `oc create job --from=cronjob/package-releases package-releases -n thoth-test-core` and I see the same behavior persist, like it will run for some time and then terminate, looks like it's not sending messages.

pacospace:
> @pacospace So, I tested by running `oc create job --from=cronjob/package-releases package-releases -n thoth-test-core` and I see the same behavior persist, like it will run for some time and then terminate, looks like it's not sending messages.

Let me have a look :)

bissenbay:
> Can you share or add more logs or inputs, verson used? :)
```
2020-09-24 13:26:45,916   1 INFO     thoth.common:325: Setting up logging to a Sentry instance 'sentry.io/1298083', environment 'ocp-test' and integrations ['SqlalchemyIntegration']
2020-09-24 13:26:45,925   1 INFO     thoth.common:368: Logging to rsyslog endpoint is turned off
2020-09-24 13:26:46,455   1 INFO     thoth.package_releases_job:54: Thoth-package-releases-job-producer v0.8.4+storages.0.25.11.common.0.19.0.messaging.0.7.6
[2020-09-24 13:26:47,153] [1] [WARNING] The database schema is not in sync with library revisions, the current library revision heads: {'6389973bd8b6'}, database heads: {'84588707e6ba'} 
[2020-09-24 13:31:47,149] [1] [WARNING] It looks like package name does not match the one parsed from artifact when parsing version from wheel - package name is tensorflow, pared version is 1.15.0rc2, artifact is tensorflow_cpu-1.15.0rc2-cp36-cp36m-manylinux2010_x86_64.whl 
[...]

PR #425, opened 2020-09-14
initial modifications for turning to producer

bissenbay:
Depends-On: https://github.com/thoth-station/package-releases-job/issues/450
Fixes: https://github.com/thoth-station/thoth-application/issues/422

### DIFF
--- a/package-releases/base/cronjob.yaml
+++ b/package-releases/base/cronjob.yaml
@@ -18,6 +18,8 @@ spec:
             - image: "package-releases-job"
               name: package-releases
               env:
+                - name: SUBCOMMAND
+                  value: "producer"
                 - name: APP_SCRIPT
                   value: "app.sh"
                 - name: KAFKA_BOOTSTRAP_SERVERS
--- a/package-releases/base/cronjob.yaml
+++ b/package-releases/base/cronjob.yaml
@@ -3,6 +3,9 @@ kind: CronJob
 apiVersion: batch/v1beta1
 metadata:
   name: package-releases
+  labels:
+    app: thoth
+    component: package-releases
 spec:
   schedule: "0 */12 * * *"
   suspend: true
@@ -13,22 +16,15 @@ spec:
   jobTemplate:
     spec:
       template:
+        metadata:
+          labels:
+            app: thoth
+            component: package-releases
         spec:
           containers:
-            - image: "package-releases-job"
+            - image: "package-releases-job:latest"
               name: package-releases
               env:
-                - name: SUBCOMMAND
-                  value: "producer"
-                - name: APP_SCRIPT
-                  value: "app.sh"
-                - name: KAFKA_BOOTSTRAP_SERVERS
-                  valueFrom:
-                    configMapKeyRef:
-                      key: kafka-bootstrap-servers
-                      name: kafka
-                - name: KAFKA_CAFILE
-                  value: "/mnt/secrets/kafka_ca.crt"
                 - name: THOTH_PACKAGE_RELEASES_ONLY_IF_PACKAGE_SEEN
                   valueFrom:
                     configMapKeyRef:
@@ -44,6 +40,11 @@ spec:
                     configMapKeyRef:
                       key: rsyslog-port
                       name: thoth
+                - name: PROMETHEUS_PUSHGATEWAY_URL
+                  valueFrom:
+                    configMapKeyRef:
+                      name: prometheus
+                      key: pushgateway-url
                 - name: SENTRY_DSN
                   valueFrom:
                     secretKeyRef:
@@ -78,10 +79,6 @@ spec:
                     secretKeyRef:
                       name: postgresql
                       key: database-name
-              volumeMounts:
-                - name: secrets
-                  mountPath: /mnt/secrets
-                  readOnly: true
               resources:
                 requests:
                   memory: "256Mi"
@@ -96,11 +93,4 @@ spec:
                 periodSeconds: 10
                 tcpSocket:
                   port: 80
-          volumes:
-            - name: secrets
-              secret:
-                secretName: "kafka"
-                items:
-                  - key: kafka_ca.crt
-                    path: kafka_ca.crt
           restartPolicy: OnFailure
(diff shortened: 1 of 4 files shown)


package-releases/base/kustomization.yaml
```diff
@@ -3,8 +3,8 @@ kind: Kustomization
 resources:
   - cronjob.yaml
   - imagestream.yaml
-generatorOptions:
-  disableNameSuffixHash: true
+  - route.yaml
+  - service.yaml
```
pacospace, line 7: Route and service are not required, we won't deploy package-release-job :) Remove also the files accordingly please

bissenbay: Thanks, force-pushed

bissenbay:
@pacospace I looked at https://github.com/thoth-station/thoth-application/blob/master/package-update/base/cronjob.yaml and that's the main difference I see that it has https://github.com/thoth-station/thoth-application/blob/master/package-update/base/cronjob.yaml#L80-L103, so I added, is there anything I should change as well?

package-releases/base/cronjob.yaml
```diff
@@ -16,13 +13,9 @@ spec:
...
     spec:
       template:
-        metadata:
-          labels:
-            app: thoth
-            component: package-releases
         spec:
           containers:
-            - image: "package-releases-job:latest"
+            - image: "package-releases-job"
```
bissenbay, line 18: Should I keep `latest` tag?

package-releases/base/kustomization.yaml
```diff
@@ -3,8 +3,6 @@ kind: Kustomization
 resources:
   - cronjob.yaml
   - imagestream.yaml
-generatorOptions:
```
bissenbay, line 6: @pacospace https://github.com/thoth-station/thoth-application/blob/master/package-update/base/kustomization.yaml doesn't contain it, so deleted here as well :smile:

package-releases/base/cronjob.yaml
pacospace: No need as it wil use the latest by default, Argo CD make sure what is in the cluster is in sync with versioning from quay :)  @harshad16 can be more verbose here :D

package-releases/base/kustomization.yaml
pacospace: Sounds good :D

bissenbay:
@pacospace `Qeb-Hwt` is failing, should I really add `.thoth.yaml`?

pacospace:
> @pacospace `Qeb-Hwt` is failing, should I really add `.thoth.yaml`?

It's meant to file in this case don't worry :) , we don't advise on repo with no python in any case and without any type of requirements

package-releases/base/cronjob.yaml
pacospace, line 18: You miss Kafka specific env variables, have a look at package update cronjob :) also how to run producer.py from app.sh :)

bissenbay: Thanks, https://github.com/thoth-station/thoth-application/pull/425/commits/33969ce76b8a00688a40479757e497120438cdfd, I think we don't provide `SUBCOMMAND` for value `producer`, as in this case we just need to run `app.sh`.

harshad16: latest is not required as @pacospace  suggested.
openshift default looks for `latest` if no tag is defined.
also, imagestreamtag in the package-release-job component would be responsible for latest tag update with new versions along with argocd.

package-releases/base/kustomization.yaml
pacospace: Actually, let's ask @harshad16, not convinced here :)

package-releases/base/cronjob.yaml
pacospace: only if there is a defaul value in `app.sh` in package release job, otherwise we have to :)

package-releases/base/kustomization.yaml
harshad16: yes, this is not needed, 
it needed when we create configmap or secret with the kustomize file, as we don't in this case, it can be removed.

package-releases/base/cronjob.yaml
bissenbay: thanks, https://github.com/thoth-station/thoth-application/pull/425/commits/45fbfc8b839c0d24951f26caffddd77854af189a

pacospace (approved):
Nice thanks!!!

pacospace:
@harshad16 @goern we are ready to test package-release-job, should we move this template to overlays/test for testing? @bissenbay

harshad16:
> @harshad16 @goern we are ready to test package-release-job, should we move this template to `overlays/test` for testing as we discussed once on the best practice? Later if all combination works we moved it to final place in `base` @bissenbay

yes, personally i feel that would be the best approach,  we could that, by doing these changes in the pr.
- Take the copy of the changed cronjob.yaml to https://github.com/thoth-station/thoth-application/package-releases/overlays/test/
- Then add the cronjob in the patch https://github.com/thoth-station/thoth-application/blob/40123270cce02cded22f6c2b1160976a559ffc4a/package-releases/overlays/test/kustomization.yaml#L9
so that kustomize patches these new change in test while deploying.

bissenbay:
Thanks @pacospace @harshad16 I pushed a new commit with those changes https://github.com/thoth-station/thoth-application/pull/425/commits/ffb3a08cc4776064d81940a26b19b6ae6de5fccb.

harshad16:
@bissenbay , please undo you change in package-releases/base/cronjob.yaml 
as it is already be covered by the test overlay cronjob.

bissenbay:
> @bissenbay , please undo your change in package-releases/base/cronjob.yaml
> as it is already be covered by the test overlay cronjob.

Thanks, https://github.com/thoth-station/thoth-application/pull/425/commits/eb2fc47c6618f84d2ffbeba0f911f25f898e0ba0

harshad16 (approved):
lgtm 
lets test this.

harshad16:
just noticed, that the cronjob is in suspended state.
it can be run as a job with following command
`oc create job --from=cronjob/package-releases package-releases -n thoth-test-core`

PR #437, opened 2020-09-16
Increase parallelism in middletier-stage

fridex:

Based on the quota increase to utilize resources in a more efficient manner.

- [x] No


fridex:
CC @harshad16

harshad16 (approved):
lgtm
bbut we would have to notice if the wc will go into oom

Issue #267, opened 2020-08-18
Converting Graph-refresh-job to a Kafka producer. 


saisankargochhayat:
As graph-refresh-job now acts as a Kafka producer and creates messages for the investigator to process, we need to change the graph-refresh-job deployment config to allow it to be a Kafka producer.
Necessary to support - https://github.com/thoth-station/graph-refresh-job/pull/448

pacospace:
Already solved by @saisankargochhayat

PR #444, opened 2020-09-16
Updated cronjob to use producer

saisankargochhayat:
To deploy new graph-refresh-job to thoth-test-core

### DIFF
--- a/graph-refresh/overlays/test/cronjob.yaml
+++ b/graph-refresh/overlays/test/cronjob.yaml
@@ -28,6 +28,8 @@ spec:
               env:
                 - name: DEBUG_LEVEL
                   value: "0"
+                - name: SUBCOMMAND
+                  value: "producer"
                 - name: APP_SCRIPT
                   value: "app.sh"
                 - name: KAFKA_BOOTSTRAP_SERVERS
@@ -41,7 +43,7 @@ spec:
                   value: "https://kubernetes.default.svc.cluster.local"
                 - name: KUBERNETES_VERIFY_TLS
                   value: "0"
-                - name: THOTH_INFRA_NAMESPACE
+                - name: NAMESPACE
                   valueFrom:
                     configMapKeyRef:
                       key: infra-namespace
--- a/graph-refresh/overlays/test/cronjob.yaml
+++ b/graph-refresh/overlays/test/cronjob.yaml
@@ -28,8 +28,6 @@ spec:
               env:
                 - name: DEBUG_LEVEL
                   value: "0"
-                - name: SUBCOMMAND
-                  value: "producer"
                 - name: APP_SCRIPT
                   value: "app.sh"
                 - name: KAFKA_BOOTSTRAP_SERVERS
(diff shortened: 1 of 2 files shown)


graph-refresh/overlays/test/cronjob.yaml
```diff
@@ -61,21 +47,21 @@ spec:
...
-                    configMapKeyRef:
-                      name: prometheus
-                      key: pushgateway-url
                 - name: SENTRY_DSN
                   valueFrom:
                     secretKeyRef:
                       name: thoth
                       key: sentry-dsn
-                - name: KNOWLEDGE_GRAPH_HOST
+                - name: THOTH_DEPLOYMENT_NAME
```
bissenbay, line 60: is this being used?

saisankargochhayat: Sorry I updated the file can you please check?

graph-refresh/overlays/test/cronjob.yaml
```diff
@@ -28,6 +28,8 @@ spec:
               env:
                 - name: DEBUG_LEVEL
                   value: "0"
+                - name: SUBCOMMAND
```
bissenbay, line 31: I think as you provided `SUBCOMMAND`, then you should add it in `app.sh` like in https://github.com/thoth-station/package-releases-job/blob/master/app.sh for https://github.com/thoth-station/graph-refresh-job/blob/master/app.sh, WDYT @pacospace ?

graph-refresh/overlays/test/cronjob.yaml
```diff
@@ -41,7 +43,7 @@ spec:
                   value: "https://kubernetes.default.svc.cluster.local"
                 - name: KUBERNETES_VERIFY_TLS
                   value: "0"
-                - name: THOTH_INFRA_NAMESPACE
+                - name: NAMESPACE
```
bissenbay, line 46: I think it's good, I see it's being used at https://github.com/thoth-station/graph-refresh-job/blob/master/producer.py#L60.

graph-refresh/overlays/test/cronjob.yaml
saisankargochhayat: Ah you are right, we dont need subcommand as graph refresh job is always a producer.

saisankargochhayat:
@harshad16 do we need anything else to close - https://github.com/thoth-station/thoth-application/issues/267
Seems like the cron-job config is correct for it to be a kafka producer

graph-refresh/overlays/test/cronjob.yaml
bissenbay: I'd add it as @pacospace advised to do so in case of `package-releases-job`, PTAL https://github.com/thoth-station/thoth-application/pull/425#discussion_r488812289.

saisankargochhayat: Hmm, the graph-refresh-jobs app.sh doesn't have the SUBCOMMAND variable  - 
https://github.com/thoth-station/graph-refresh-job/blob/28998f7d6e4b113206cdd8c040c1fcdd6d803a3b/app.sh#L20

harshad16 (approved):
lgtm

PR #449, opened 2020-09-17
Increase parallelism in middletier even more

fridex:
Feels like we are utilizing 60 - 75% now.

- [x] No


harshad16 (approved):
ok, lets test this out.
lets keep an eye out for wc running into oom.

PR #454, opened 2020-09-17
Add thoth config

saisankargochhayat:
Add thoth config to support version releases

### DIFF
--- a/.thoth.yaml
+++ b/.thoth.yaml
@@ -1,16 +1,3 @@
-host: khemenu.thoth-station.ninja
-tls_verify: true
-requirements_format: pipenv
-
-runtime_environments:
-  - name: 'rhel:8'
-    operating_system:
-      name: rhel
-      version: "8.0"
-    python_version: "3.6"
-    recommendation_type: latest
-    limit_latest_versions: 5
-
 managers:
   - name: version
     configuration:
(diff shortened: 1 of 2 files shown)


.thoth.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+runtime_environments:
+  - name: 'rhel:8'
+    operating_system:
+      name: rhel
+      version: "8.0"
+    python_version: "3.6"
+    recommendation_type: latest
+    limit_latest_versions: 5
+
```
harshad16, line 13: we dont need this here as well, right?

version.py
```diff
@@ -0,0 +1,4 @@
+"""Thoth Application."""
+
+__name__ = "kebechet"
```
harshad16, line 3:
```suggestion
__name__ = "thoth-station"
```
or something else?

.thoth.yaml
saisankargochhayat: We can remove these

harshad16 (approved):
Thanks :100:

.thoth.yaml
fridex: Yes, `limit_latest_versions: 5` is no longer supported. If you spot it used somewhere, please remove it from `.thoth.yaml`.

PR #472, opened 2020-09-23
Propagate deployment-name and Amun API url to Amun deploymentconfig

fridex:

- [x] No

### DIFF
--- a/amun/base/deploymentconfig.yaml
+++ b/amun/base/deploymentconfig.yaml
@@ -33,6 +33,11 @@ spec:
               value: "0"
             - name: THOTH_SENTRY_IGNORE_LOGGER
               value: "connexion.decorators.validation"
+            - name: THOTH_AMUN_API_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: amun
+                  key: amun-api-url
             - name: THOTH_AMUN_INSPECTION_NAMESPACE
               valueFrom:
                 configMapKeyRef:


amun/base/deploymentconfig.yaml
```diff
@@ -33,6 +33,16 @@ spec:
...
+            - name: THOTH_AMUN_API_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: amun
+                  key: amun-api-url
+            - name: THOTH_AMUN_DEPLOYMENT_NAME
+              valueFrom:
+                configMapKeyRef:
+                  name: amun
+                  key: deployment-name
```
harshad16, line 45: is this a replacement for this https://github.com/thoth-station/thoth-application/blob/e478b846cf6242b105aacc634e8147ca077c253e/amun/base/deploymentconfig.yaml#L91
or both are to be provided ?

fridex: Good catch! :+1: I missed that one. Adjusted.

harshad16: can you point me to the code where all these would be used by, i m kinda forgetting how this would be used in whole process.
Thanks.

harshad16: ah i see it here: https://github.com/thoth-station/amun-api/pull/567/files
thanks

harshad16 (approved):
lgtm 
thanks

amun/base/deploymentconfig.yaml
fridex: It's propagated to the specification document stored on Ceph: https://github.com/thoth-station/amun-api/pull/567/files This specification document specifies the whole request to Amun and states all the input with some metadata. As we want to notify that an inspection finished via a Kafka message, we need to propagate this information in the message somewhere:

https://github.com/thoth-station/messaging/pull/270

It would be also good to have this stated in the specification document if we plan to have multiple Amun deployments (multiple clusters) to aggregate results.

PR #477, opened 2020-09-23
Minor adjustments and add metrics for qeb-hwt and kebechet

pacospace:

Requirements for metrics > 20 to be scrapred by Prometheus ( that scrapes every 20 seconds ), with 25 seconds we should be fine.


harshad16 (approved):
lgtm

Issue #398, opened 2020-09-10
Amun workflow can send message to Kafka to schedule graph-sync. [8pt]


pacospace:
As User of Amun,

I would like to have the option to send a message to Kafka to know that an inspection ended succesfully, so that a consumer of that message can take action on that.

- [x] Create in thoth-messaging message `InspectionRunMessage` with document ID regarding inspection run just successfully concluded.
- [x] Release new thoth-messaging
- [ ] Create task in workflow-helpers to parse inputs for `InspectionRunMessage`

As Thoth User of Amun,

Beside steps above I want to use that message in investigator to schedule graph-sync job so that Thoth can learn about inspections.

- [x] Create a `graph-sync Workflow` with one task and add it to thoth-application
- [x] Create logic in thoth-common to schedule the `graph-sync workflow`
- [x] Release new thoth-common
- [ ] Create logic in investigator to consume `InspectionRunMessage` and schedule graph-sync Workflow

Add to amun workflow:

- [ ] Make sure amun has thoth-messaging and workflow-helpers images available.
- [ ] Add tasks to amun template:
- parse inspection inputs
- send message using thoth-messaging CLI

with condition based on env variable to decide if the message need to be sent or not: `THOTH_AMUN_SEND_MESSAGE`

Reference discussion: https://github.com/thoth-station/thoth-application/issues/352

KPostOffice:
What should the contents of `InspectionRunMessage` be?
and also I don't understand why we can't directly use GraphSync using the same conditional logic.
@fridex @goern

fridex:
> What should the contents of `InspectionRunMessage` be?

The most important part is inspection_id so we know what inspection is finished.

> and also I don't understand why we can't directly use GraphSync using the same conditional logic.
> @fridex @goern

Amun is a standalone component - we want to deploy it in different clusters with different HW to obtain inspection results. Then the message will notify a Thoth deployment that an inspection can be obtained and synced.

fridex:
This is duplicate of https://github.com/thoth-station/amun-api/issues/552

PR #480, opened 2020-09-23
add new workflow for graph-sync

KPostOffice:

https://github.com/thoth-station/thoth-application/issues/398

- [ ] Yes
- [x] No

## Test or Stage Environment

We want to be able to schedule graph-sync independently so that we can trigger it using thoth-common

## Description

I added a new workflow template which takes the document ID and force sync as parameters and works the exact same as the current template save the fact that it pulls the document from ceph rather than from a local directory.

I then also added a openshift template that will be used in thoth.common to schedule graph syncs specifically when triggered by amun `InspectionCompleted` messages


pacospace (commented):
lgtm

KPostOffice:
@goern

harshad16 (approved):
thanks, let's get this in.

PR #481, opened 2020-09-23
delete pods to free PVC

KPostOffice:
Delete pods before workflow to free PVC

### DIFF
--- a/security-indicators/base/openshift-templates/security-indicators.yaml
+++ b/security-indicators/base/openshift-templates/security-indicators.yaml
@@ -31,6 +31,8 @@ objects:
         secondsAfterCompletion: 25
         secondsAfterSuccess: 25
         secondsAfterFailure: 25
+      podGC:
+        strategy: OnPodCompletion
       arguments:
         parameters:
           - name: THOTH_SECURITY_INDICATORS_JOB_ID
@@ -85,7 +87,6 @@ objects:
                 - 900
               value: "{{workflow.duration}}"
 
-      onExit: exit-handler
       serviceAccountName: argo
       templates:
         - name: security-indicators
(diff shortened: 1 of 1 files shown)


security-indicators/base/openshift-templates/security-indicators.yaml
```diff
@@ -85,7 +87,6 @@ objects:
                 - 900
               value: "{{workflow.duration}}"
 
-      onExit: exit-handler
```
pacospace, line 88: isn't this required to remove the PVC? removing the pod is going to delete also the `PersistentVolumeClaim` object?

goern: Deleting the pod will also delete its log, do we need that log?

goern: > isn't this required to remove the PVC? removing the pod is going to delete also the `PersistentVolumeClaim` object?

That's a question about Argo Workflows behavior, for OpenShift only the answer is 'no'

pacospace: That is what I thought and I think argo cannot do that, podGC is for pods not for PVC.

pacospace: > Deleting the pod will also delete its log, do we need that log?

We already keep the life of workflow to 25 seconds in middletier, the logs should be stored on Ceph if we need them, we need to set that in the workflow-controller CM

KPostOffice: It's not required to remove the PVC, all that is required is that the pods be deleted

pacospace: and who removes the PVC requested? we need to free it

KPostOffice: It is removed automatically once all pods have been deleted

pacospace: I didn't know, then I trust you :)

fridex: I think it is deleted based on the owner reference configured by argo.

Speaking of this, It would be great to check the reclaim policy for volumes in the cluster - for some configuration, we can eat up all the available volumes if they are not recycled. I believe we are not affected, we would already encounter this with the number of workflows we processed.

goern:
is this still valid?

goern:
@KPostOffice @pacospace is this still valid?

pacospace:
After discussing with @KPostOffice, it can be closed. He mentioned ttl strategy is working well already

PR #487, opened 2020-09-24
Use correct URL after migration to OCP

pacospace:


harshad16:
i was just about to open the pr. 
i was thinking to remove the configmap from adviser itself. 
and make the trigger-integration key as qeb-hwt-workflow.

PR #491, opened 2020-09-24
include production manifest files

harshad16:
include production manifest files

## Environment

- [x] Pull Requests added content to the Prod environment


fridex (commented):
:eyes: :partying_face:

goern:
Shall we call the overlays for production "ocp4-psi" and "moc-kaizen"? Would we have deployment_names with the same values?

harshad16:
> Shall we call the overlays for production "ocp4-psi" and "moc-kaizen"? Would we have deployment_names with the same values?
Good questions, we could have these name for the overlays.

harshad16:
this is ready for merging 
if we would like to change in the overlays names, i have no personal objection to it.
@goern  @fridex  @pacospace

fridex:
> if we would like to change in the overlays names, i have no personal objection to it.

Do we want to have one just one "`prod`" or do we plan to differentiate more?

fridex:
> > if we would like to change in the overlays names, i have no personal objection to it.
> 
> Do we want to have one just one "`prod`" or do we plan to differentiate more?

This was also to @goern comment in https://github.com/thoth-station/thoth-application/pull/491#issuecomment-699961871. What I meant - one "prod" deployment and others would be prod-like deployments in different clusters?

harshad16:
> This was also to @goern comment in [#491 (comment)](https://github.com/thoth-station/thoth-application/pull/491#issuecomment-699961871). What I meant - one "prod" deployment and others would be prod-like deployments in different clusters?

production environment, sharing similar type of cluster based evnironments(ex, ceph, prometheus) can still use one overlays.
we could just have different overalys in core and amun, as they have the configmap and secrets sets for deployment name as such.

core overlays - ocp-prod, ocp4-prod, ...
other application - prod.

however when we change cluster, we have to handle rolebinding based on cluster requirements, some application has rolebinding for clusterroles which need modification.


Either we can have different overlays in all apps for consistency, or different overlays only in required applications like(core, amun, etc)

PR #499, opened 2020-09-25
bump up the core images for stage

harshad16:
bump up the core images for stage

### DIFF
--- a/core/overlays/stage/imagestreamtags.yaml
+++ b/core/overlays/stage/imagestreamtags.yaml
@@ -1,17 +1,16 @@
----
 apiVersion: image.openshift.io/v1
 kind: ImageStream
 metadata:
   name: thoth-toolbox
 spec:
   tags:
-    - name: latest
-      from:
-        kind: DockerImage
-        name: "quay.io/thoth-station/thoth-toolbox:v0.2.0"
-      importPolicy: {}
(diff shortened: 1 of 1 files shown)


core/overlays/stage/imagestreamtags.yaml
```diff
@@ -1,70 +1,69 @@
...
 spec:
   tags:
-    - name: latest
-      from:
-        kind: DockerImage
-        name: "quay.io/thoth-station/thoth-toolbox:v0.2.0"
-      importPolicy: {}
-      referencePolicy:
-        type: Local
+  - name: latest
```
pacospace, line 7: is the indentation correct?

pacospace (commented):
one question otherwise lgtm

core/overlays/stage/imagestreamtags.yaml
harshad16: yes, it works both ways (like and with one more tab space)

harshad16: the imagestream in test get indented like this as it update by sesheta , so just followed it

pacospace: we follow :robot:

pacospace (approved):

PR #529, opened 2020-10-02
Deactivate SI scheduling from graph-refresh as investigator will schedule SI analysis once package is solved

pacospace:


goern:
Shall we externalize these configurables into a configmap? :)

pacospace:
> Shall we externalize these configurables into a configmap? :)

Why in CM?

PR #539, opened 2020-10-05
Bump investigator to v0.5.3

pacospace:


goern:
@pacospace @bissenbay can we think about a few tests that can prove investigator is going good, some integration tests?

/approve

PR #576, opened 2020-10-20
Increase ingestion

pacospace:

Increase ingestion so that Thoth can continuously learn.

Screenshot from 2020-10-20 08-22-57


harshad16 (approved):
:+1:

PR #600, opened 2020-10-30
Increase packages considered while we have 1 solver running

pacospace:

There are moments where Thoth is not learning. Let's increase while we have 1 solver to learn more.

Screenshot from 2020-10-30 09-13-23


harshad16 (approved):
lgtm :+1:

PR #606, opened 2020-11-03
Add additional wait before the resolver is terminated

fridex:

Related: https://github.com/thoth-station/adviser/pull/1527

- [x] No

With  https://github.com/thoth-station/adviser/pull/1527 we implemented https://github.com/thoth-station/adviser/issues/1526. When running adviser in deployment, we expect the first liveness probe to send SIGUSR1 signal which tells predictor to perform the exploitation phase as it will soon be terminated. The second signal SIGINT is handled by the resolver to stop the resolution process and report back any results computed so far. With SIGUSR1 and subsequent SIGINT signals, we are able to start the exploitation phase (if it haven't been reached until that point) before the adviser terminates and deliver better results that were found until that point.

### DIFF
--- a/adviser/base/argo-workflows/advise.yaml
+++ b/adviser/base/argo-workflows/advise.yaml
@@ -189,6 +189,7 @@ spec:
             command:
               - python3
               - liveness.py
-          failureThreshold: 1
+          failureThreshold: 3
           initialDelaySeconds: 1500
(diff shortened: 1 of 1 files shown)


harshad16 (approved):
lgtm :+1

Thanks for teh explanation.

PR #579, opened 2020-10-20
deployment manifest for ocp4 stage deployment

harshad16:
## Related Issues and Dependencies

Related-to: #504 

## Description

deployment manifest for ocp4 stage deployment
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

PR #624, opened 2020-11-09
Use here-strings instead of here documents

fridex:

Fixes: #579

- [x] No


harshad16 (approved):
lgtm

PR #625, opened 2020-11-09
Fix expansion of env variables

fridex:
## Related Issues and Dependencies

Related: #624 

```
+ results=/mnt/build/
+ mkdir -p /mnt/build/
+ kubectl logs inspection-test-dca0953e-1-build
+ jq .
+ cat registry.access.redhat.com/ubi8/python-36, batch_size: 1, build: '{requests:' '{cpu:' 1, hardware: '{cpu_family:' 6, cpu_model: 94, physical_cpus: 32, processor: 'Intel-Xeon-Processor-Skylake-IBRS},' memory: '1Gi}},' environment: '[{name:' string, value: 'string}],' files: '[{content:' string, path: 'string}],' identifier: test, package_manager: micropipenv, packages: '[vim],' python: '{requirements:' '{},' requirements_locked: '{}},' python_packages: '[pipenv],' run: '{requests:' '{cpu:' 1, hardware: '{cpu_family:' 6, cpu_model: 94, physical_cpus: 32, processor: 'Intel-Xeon-Processor-Skylake-IBRS},' memory: '512Mi}},' script: '#!/usr/bin/bash\necho "Here should be run tests..."\n,' update: true, upgrade_pip: false, @created: '2020-11-09T12:27:52.078021}'
cat: registry.access.redhat.com/ubi8/python-36,: No such file or directory
cat: 'batch_size:': No such file or directory
cat: 1,: No such file or directory
cat: 'build:': No such file or directory
cat: '{requests:': No such file or directory
cat: '{cpu:': No such file or directory
cat: 1,: No such file or directory
cat: 'hardware:': No such file or directory
cat: '{cpu_family:': No such file or directory
cat: 6,: No such file or directory
cat: 'cpu_model:': No such file or directory
cat: 94,: No such file or directory
```

## This introduces a breaking change

- [x] No

PR #626, opened 2020-11-09
Another try to fix expansion

fridex:

```
+ results=/mnt/build/
+ mkdir -p /mnt/build/
+ kubectl logs inspection-test-8641726c-1-build
+ cat '"Here' should be run 'tests..."n, update: true, upgrade_pip: false, @created: 2020-11-09T12:54:44.755753}'
+ jq .
cat: '"Here': No such file or directory
cat: should: No such file or directory
cat: be: No such file or directory
cat: run: Is a directory
cat: 'tests..."n, update: true, upgrade_pip: false, @created: 2020-11-09T12:54:44.755753}': No such file or directory
```

Related: #624 #625 

- [x] No

### DIFF
--- a/amun/base/argo-workflows/inspection-build-with-cpu.yaml
+++ b/amun/base/argo-workflows/inspection-build-with-cpu.yaml
@@ -267,10 +267,18 @@ spec:
           mkdir -p "$results"
           kubectl logs "{{inputs.parameters.inspection-id}}-1-build" > "${results}/log"
 
-          cat <<< "{{inputs.parameters.specification}}" | jq . > "${results}/specification"
+          spec=`cat <<__EOF__
+          {{inputs.parameters.specification}}
+          __EOF__
+          `
+          echo "${spec}" | jq . > "${results}/specification"
 
           set +e
-          cat <<< "{{inputs.parameters.dockerfile}}" > "${results}/Dockerfile"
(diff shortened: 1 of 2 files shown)


fridex:
... so it looks like a simple string replace.

amun/base/argo-workflows/inspection-build-with-cpu.yaml
```diff
@@ -267,10 +267,18 @@ spec:
           mkdir -p "$results"
           kubectl logs "{{inputs.parameters.inspection-id}}-1-build" > "${results}/log"
 
-          cat <<< "{{inputs.parameters.specification}}" | jq . > "${results}/specification"
+          spec=`cat <<__EOF__
```
harshad16, line 270: can we not use direct echo ?
```suggestion
       echo "{{inputs.parameters.specification}}" | jq . > "${results}/specification"
```

fridex: We cannot. `{{inputs.parameters.specification}}` part is simply replaced with the parameter value and this causes issues with matching " (it holds a JSON).

harshad16: ok thanks, lets get this in.

harshad16 (approved):
thanks

PR #648, opened 2020-11-13
Decrease beam width to 10000

fridex:
... it still feels like keeping 10,000 candidates is more than enough.

- [ ] Yes
- [x] No


harshad16 (approved):
lgtm

PR #653, opened 2020-11-16
set other solvers to be solved

harshad16:
set other solvers to be solved

### DIFF
--- a/core/overlays/stage/common/configmaps.yaml
+++ b/core/overlays/stage/common/configmaps.yaml
@@ -37,6 +37,8 @@ metadata:
 data:
   solvers: |
     solver-rhel-8-py36
+    solver-fedora-32-py37
+    solver-fedora-32-py38
 ---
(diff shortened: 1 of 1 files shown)


core/overlays/stage/common/configmaps.yaml
```diff
@@ -37,6 +37,10 @@ metadata:
 data:
   solvers: |
     solver-rhel-8-py36
+    solver-fedora-31-py38
+    solver-fedora-31-py37
```
fridex, line 41: Do we want to invest time to Fedora 31? I think we could include F33 instead.

fridex: https://endoflife.software/operating-systems/linux/fedora

harshad16: ack, will remove the f31.

fridex (approved):
:100:

PR #670, opened 2020-11-20
Reduce workflows to avoid Kafka limit for messages (Data Driven Engineering)

pacospace:

Fixes: https://github.com/thoth-station/thoth-application/issues/602

Reason:
Screenshot from 2020-11-20 20-40-54

### DIFF
--- a/graph-refresh/base/cronjob.yaml
+++ b/graph-refresh/base/cronjob.yaml
@@ -63,7 +63,7 @@ spec:
                 - name: THOTH_GRAPH_REFRESH_SECURITY
                   value: "1"
                 - name: THOTH_GRAPH_REFRESH_COUNT
-                  value: "400"  # TODO: Reduce when more solvers are running
+                  value: "200"
                 - name: THOTH_DEPLOYMENT_NAME
                   valueFrom:
                     configMapKeyRef:


KPostOffice:
@pacospace I think this is happening because of a missing image

pacospace:
> @pacospace I think this is happening because of a missing image

so shall we remove the solver from solver list? so at least the system keeps learning until the image is created.. I don't see it on quay. WDYT?

pacospace:
> > @pacospace I think this is happening because of a missing image
> 
> so shall we remove the solver from solver list? so at least the system keeps learning until the image is created.. I don't see it on quay. WDYT?

@KPostOffice all image for solver seems to be present actually: https://github.com/thoth-station/thoth-application/blob/4ac9703cee0da6b96657486925988f0ff4f47fb1/core/overlays/stage/common/configmaps.yaml#L38, not sure where `ubi-8-py38` is coming from

goern:
what do we do with this PR?

pacospace:
> what do we do with this PR?

We can close it for now, it appears once removed those workflows all went back to normal.

harshad16:
The graph-refresh run with additional solvers is overwhelming the workflow in middletier-stage.
graph-refresh

will be merging this, please feel free to open an issue. 
/approve

pacospace:
> The graph-refresh run with additional solvers is overwhelming the workflow in middletier-stage.
> graph-refresh
> 
> will be merging this, please feel free to open an issue.
> /approve

that graph is from ocp4-stage but this PR reduced the one for ocp-stage, we need to open new one for that :)

harshad16:
> that graph is from ocp4-stage but this PR reduced the one for ocp-stage, we need to open new one for that :)
as the change is in the base, this would reduce workflows accordingly in both clusters.

pacospace:
> > that graph is from ocp4-stage but this PR reduced the one for ocp-stage, we need to open new one for that :)
> 
> as the change is in the base, this would reduce workflows accordingly in both clusters.

right, thanks Harsad!! :rocket:

PR #671, opened 2020-11-23
Reduce time after which adviser is cut off

fridex:

Reduce time for adviser runs, at least for now so we get results sooner.


harshad16 (approved):
thanks :100:

PR #692, opened 2020-12-01
Add kebechet workflow, remove old processing step

xtuchyna:
https://github.com/thoth-station/mi-scheduler/pull/58

- [ ] Yes
- [X] No

## Test or Stage Environment

## Description
New `mi-kebechet-analyse-workflow`
Removes old processing step of mi workflow
Old `mi-worfklow` was renamed to `mi-anaylse-workflow`

### DIFF
--- a/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
+++ b/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
@@ -9,8 +9,6 @@ metadata:
 
 spec:
   serviceAccountName: argo
-  entrypoint: mi-workflow
-
   templates:
     - name: mi-anaylse-workflow
       resubmitPendingPods: true
@@ -32,7 +30,7 @@ spec:
           - name: REPOSITORY
       steps:
         - - name: collect-kebechet-knowledge
-            template: kebechet
+            template: kebechet-analyse
             arguments:
               parameters:
                 - name: REPOSITORY
@@ -81,7 +79,7 @@ spec:
           - name: PYTHONPATH
             value: "."
 
-    - name: kebechet
+    - name: kebechet-analyse
       resubmitPendingPods: true
       inputs:
         parameters:
--- a/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
+++ b/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
@@ -12,7 +12,7 @@ spec:
   entrypoint: mi-workflow
 
   templates:
-    - name: mi-workflow
+    - name: mi-anaylse-workflow
       resubmitPendingPods: true
       inputs:
         parameters:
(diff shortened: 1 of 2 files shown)


mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
```diff
@@ -9,10 +9,8 @@ metadata:
 
 spec:
   serviceAccountName: argo
-  entrypoint: mi-workflow
-
   templates:
-    - name: mi-workflow
+    - name: mi-anaylse-workflow
```
pacospace, line 13: don't we keep two templates? One for general MI to collect all knowledge and one specific for kebechet (subset)?

xtuchyna: yes that's right, gonna fix it

xtuchyna: okay @pacospace so in the end the base `mi` workflow was extended with `ENTITIES` parameter that specifies which entities would be analysed. I've removed the kebechet workflow because it probably should not be here, it is redundant, and we can genrally make the same call from `thoth.common` with just specified parameter :)

xtuchyna:
hello hello @pacospace

mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
```diff
@@ -9,81 +9,34 @@ metadata:
 
 spec:
   serviceAccountName: argo
-  entrypoint: mi-workflow
```
pacospace, line 12: why removed?

mi-scheduler/base/openshift-templates/mi-run-template.yaml
```diff
@@ -13,6 +13,11 @@ parameters:
     description: GitHub repository whose Health Indicator will be computed
     displayName: inspected GitHub repository
 
+  - name: ENTITIES
+    required: true
```
pacospace, line 17: shouldn't be optional? If it is not provided, all entities are considered

pacospace (commented):
Two questions :)

mi-scheduler/base/openshift-templates/mi-run-template.yaml
xtuchyna: yes, that's actually true

xtuchyna: ugh, the remaining ``` lasted, weird. Consider this fixed

mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
xtuchyna: forgot about that one from previously, thx!

pacospace (approved):
Thanks :D

PR #704, opened 2020-12-04
Bumped kebechet in test

saisankargochhayat:
Wait till quay gets image - https://quay.io/repository/thoth-station/kebechet-job?tab=tags
Related - https://github.com/thoth-station/kebechet/pull/649

### DIFF
--- a/kebechet/overlays/test/imagestreamtag.yaml
+++ b/kebechet/overlays/test/imagestreamtag.yaml
@@ -8,7 +8,7 @@ spec:
     - name: latest
       from:
         kind: DockerImage
-        name: quay.io/thoth-station/kebechet-job:v1.2.1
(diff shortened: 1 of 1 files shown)


saisankargochhayat:
@harshad16 could you please look at this image not being released?

pacospace (commented):
isn't test supposed to be updated automatically?

pacospace:
> > isn't test supposed to be updated automatically?
> 
> It stays the same even if the image is pushed, is there a way to do it automatically?

No I mean Tekton should do that, @harshad16 right? 

I can approve it for now, but maybe better to check why that is not done automatically also

harshad16:
This is not needed, test imagestream are automatically set by the pipeline,

harshad16:
/unhold
there seems to be issue from pipeline side.
the image is however available, i will merge this. 
/approve

PR #705, opened 2020-12-04
fix the referencing to the correct image for kebechet workflow

harshad16:
fix the referencing to the correct image for kebechet workflow


pacospace (approved):
:rocket: Thanks!!!!

PR #782, opened 2021-01-21
Bump graph refresh v0.3.1

pacospace:


graph-refresh/base/cronjob.yaml
```diff
@@ -46,11 +46,6 @@ spec:
                   value: "https://kubernetes.default.svc.cluster.local"
                 - name: KUBERNETES_VERIFY_TLS
                   value: "0"
-                - name: NAMESPACE
```
harshad16, line 49: why is the namespace env var removed?

pacospace: It was not required, we removed it from graph-refresh, see: https://github.com/thoth-station/graph-refresh-job/pull/545.

Thanks for review, I realized as I'm changing base, this will affect also prod env, should I create patches for cronjob.yaml only for test and stage? Or easier just to add a TODO to remove this unused env variable once prod will resemble stage and leave that env variable for now? 2 option also won't affect test and stage.

wdyt?

harshad16: I think we should go with option 2 as you suggested :)

harshad16 (approved):
lgtm, thanks for the changes :100:

PR #794, opened 2021-01-26
Bump components versions to stage

pacospace:
Bump components to stage with new metric feature for metrics-exporter, investigator and management-api


fridex (approved):
:100: :+1: :ship:

PR #804, opened 2021-01-26
use a postsync hook job for deployment trigger

harshad16:
use a postsync hook job for deployment trigger

## Description

https://github.com/AICoE/aicoe-cd/issues/233#issuecomment-767672303
we can test if resource hook can help us with triggering deploymentconfig.

### DIFF
--- a/user-api/overlays/test/deployment-trigger.yaml
+++ b/user-api/overlays/test/deployment-trigger.yaml
@@ -1,0 +1,21 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: deployment-trigger
+  annotations:
+    argocd.argoproj.io/hook: PostSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      containers:
+        - name: deployment-trigger
+          image: quay.io/openshift/origin-cli:latest
+          command:
+            - "oc"
+            - "rollout"
+            - "latest"
+            - "user-api"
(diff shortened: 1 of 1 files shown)


user-api/overlays/test/deployment-trigger.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: deployment-trigger
+  annotations:
+    argocd.argoproj.io/hook: PostSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      containers:
+        - name: chat-notification
```
goern, line 13: this should be changed for consistency

user-api/overlays/test/deployment-trigger.yaml
```diff
@@ -0,0 +1,21 @@
...
+          image: quay.io/openshift/origin-cli:latest
+          command:
+            - "oc"
```
goern, line 16: is this automatically authenticated as the default user of which namespace?!

goern (changes requested):

user-api/overlays/test/deployment-trigger.yaml
harshad16: not entirely sure about this, would like to test it out.

user-api/overlays/test/deployment-trigger.yaml
harshad16: thanks

PR #852, opened 2021-02-03
 bump the version of graph-sycn in stage 

harshad16:
bump the version of graph-sycn in stage 
Already tested in the test env.


pacospace (approved):
Thanks!! 💯

pacospace:
> bump the version of graph-sycn in stage
> Already tested in the test env.

what about the messaging library? @harshad16

PR #894, opened 2021-02-12
Add integration tests

fridex:

Depends-On: https://github.com/thoth-station/integration-tests/pull/109

- [x] No

With this change, we will be able to deploy integration tests as an s2i application run as a cronjob periodically. We can run it inside any namespace in the internal network (as integration tests send email to mailing lists) to verify the application works correctly. The cronjob can be run each per deployment we want to run integration tests against.

### DIFF
--- a/integration-tests/base/cronjob.yaml
+++ b/integration-tests/base/cronjob.yaml
@@ -1,0 +1,65 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: integration-tests
+  labels:
+    app: thoth
+    component: integration-tests
+spec:
+  schedule: "@daily"
+  suspend: true
+  successfulJobsHistoryLimit: 5
+  failedJobsHistoryLimit: 4
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: null
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          labels:
+            app: thoth
+            component: integration-tests
+        spec:
+          restartPolicy: OnFailure
+          containers:
+            - image: "integration-tests:latest"
+              name: integration-tests
+              env:
+                - name: SEND_EMAIL
+                  valueFrom:
+                    configMapKeyRef:
+                      key: send-mail
+                      name: integration-tests
+                - name: THOTH_DEPLOYMENT_NAME
+                  valueFrom:
+                    configMapKeyRef:
+                      key: deployment-name
+                      name: integration-tests
+                - name: THOTH_USER_API_HOST
+                  valueFrom:
+                    configMapKeyRef:
+                      key: user-api
+                      name: integration-tests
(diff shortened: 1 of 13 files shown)


integration-tests/base/cronjob.yaml
```diff
@@ -0,0 +1,47 @@
...
+                - name: THOTH_DEPLOYMENT_NAME
+                  value: ""
+                - name: THOTH_USER_API_HOST
+                  value: ""
+                - name: THOTH_MANAGEMENT_API_HOST
+                  value: ""
+                - name: THOTH_AMUN_API_HOST
+                  value: ""
+                - name: THOTH_MANAGEMENT_API_SECRET
+                  value: ""
```
fridex, line 40: Rather than having these explicitly stated for each cronjob separately (and maintain multiple cronjobs), we could create overlays that setup environment variables for specific deployments tested.

@harshad16 @pacospace wdyt?

goern: this cronjob seems to be specific to ocp4-stage anyways, so we should put it in that overlay?

fridex: I was wondering if we want to have integration tests running also for prod deployment besides stage. We can mark only some tests to be run specifically for selected deployments - just to make sure if the deployment is in a healthy state.

goern: integration tests are running on moc every 8 hours... https://prow.operate-first.cloud/view/s3/ci-prow-artifacts--5de4780e-313e-46c0-8758-209adf0ea179/prow-logs/logs/integration-tests-moc/1360107489953058816
which is configured via https://github.com/thoth-station/thoth-application/blob/master/prow/overlays/cnv-prod/config.yaml#L299-L313

goern: we could even upload the generated html page, so that it would be an artifact vivible at https://prow.operate-first.cloud/view/s3/ci-prow-artifacts--5de4780e-313e-46c0-8758-209adf0ea179/prow-logs/logs/integration-tests-moc/1360107489953058816

fridex: I'm fine with both solutions - originally I wanted to have e-mails (so it's a push rather than pull for having test results). If we are fine with maintaining also prow then I will rework this just for the stage environment.

harshad16: @fridex , we have configmaps for different environments, in [core/overlays/{env}/common/configmap] (https://github.com/thoth-station/thoth-application/blob/master/core/overlays/cnv-prod/common/configmaps.yaml)
in this way we dont do multiple cronjob configuration

harshad16 (approved):
/unhold
thanks for all the addition :100:

fridex:
> /unhold
> thanks for all the addition

Thanks for the explanation! :+1:

PR #898, opened 2021-02-12
switch on all solver in the stage env

harshad16:
switch on all solver in the stage env


fridex (approved):
lgtm :+1:

Issue #902, opened 2021-02-12
Advise workflow fails if software stack fails to be propagated via env variables


fridex:
**Describe the bug**

When asking for an advise with a large stack, advise workflow does not start at all. It looks like the main reason is Pipfile.lock stored in an environment variable.

**To Reproduce**
1. Clone https://github.com/thoth-station/elyra-aidevsecops-tutorial/
2. Run `thamos advise` and let the user stack be submitted
3. See an error

**Expected behavior**

The advise should be computed.

**Additional context**

As we store the advise request on Ceph, we can extract inputs to adviser from the request. This way it does not need to be propagated via environment variables.

goern:
cant we transfer the info via a message's content?

fridex:
> cant we transfer the info via a message's content?

 Messages content is translated to environment variables now.

We store the request payload with Pipfile/Pipfile.lock and library usage on ceph:

https://github.com/thoth-station/user-api/blob/b0b72ace062598172d29b07d22d6d74a459570c1/thoth/user_api/api_v1.py#L381-L386

so we could provide them to the adviser container using Argo's input artifacts. This should be scalable and straightforward to implement with the current implementation - just point adviser container to the downloaded artifacts instead of passing the payload directly via env vars.

fridex:
/assign @harshad16 

I guess it is not necessary to include this in the current release cycle. Up to you, the work should be done.

PR #912, opened 2021-02-15
WIP: Use prepared data from Ceph do not propagate large data in env vars

fridex:

Fixes: https://github.com/thoth-station/thoth-application/issues/902
- [x] Related: https://github.com/thoth-station/common/pull/1109
- [x] Related: https://github.com/thoth-station/adviser/pull/1670
- [x] Related: https://github.com/thoth-station/management-api/pull/705
- [ ] Related: https://github.com/thoth-station/investigator/pull/445
- [x] Related: https://github.com/thoth-station/user-api/pull/1259
- [x] Related: https://github.com/thoth-station/storages/pull/2221
- [x] Related: https://github.com/thoth-station/management-api/pull/707
- [x] Related: https://github.com/thoth-station/messaging/pull/359

- [x] Yes

As this is a breaking change, I'm not sure if we want to adjust directly base templates for all the deployments since this will require updating adviser in all deployments. We can probably start with the test environment (copy the templates there) and start rolling to stage and prod subsequently.

CC @harshad16 @pacospace wdyt?


harshad16 (commented):
which version of adviser support these changes?

fridex:
> which version of adviser support these changes?

Not release yet. The linked PRs need to go in first (with corresponding releases).

fridex:
> Earlier missed to see the question in the description.
> so as we want to test this in test,
> we need to move all these changes to test overlay.
> 
> simple method, copy over the both sub directories to test overlay.
> then add them to patchstrategymerge.
> that should do the trick

OK, so let's do it in a non-destrictive way. Let's keep the message data in the messages for now and at the same time let's configure adviser to *optionally* provide prepared data. If a flag is turned on, it will fallback to the legacy version so we do not break the system. Also, let's include your suggested change in the manifests. I'll adjust PRs accordingly tomorrow. This way we can fallback to old implementation if anything goes wrong.

fridex:
Replaced with https://github.com/thoth-station/thoth-application/pull/982

PR #916, opened 2021-02-15
Validate responses by Management API in stage and test environments

fridex:

Related: https://github.com/thoth-station/management-api/pull/704 (needs a release)

- [x] No

### DIFF
--- a/management-api/base/deploymentconfig.yaml
+++ b/management-api/base/deploymentconfig.yaml
@@ -34,8 +34,8 @@ spec:
             - name: THOTH_API_VALIDATE_RESPONSES
               valueFrom:
                 configMapKeyRef:
-                  key: management-api
-                  name: thoth
+                  key: validate-responses
+                  name: management-api
             - name: AMUN_API_URL
               valueFrom:
                 configMapKeyRef:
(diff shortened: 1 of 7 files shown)


management-api/base/deploymentconfig.yaml
```diff
@@ -31,6 +31,11 @@ spec:
               value: "0"
             - name: THOTH_MANAGEMENT_API_RUN_MIGRATIONS
               value: "0"
+            - name: THOTH_API_VALIDATE_RESPONSES
+              valueFrom:
+                configMapKeyRef:
+                  key: management-api
+                  name: thoth
```
goern, line 38: will this irritate cnv-prod? as the configmap does not have the required key?! @harshad16

fridex: I put `validate-responses` to the base configmap which is not patched for prod deployment. But maybe I did something wrong?

harshad16: I think , this is good enough. As it an environment variable , so even if it is present and the management API version is not consuming it, it would not break the deployment. 
however the envvar referencing seems to be misconfigured in here.

harshad16, line 38:
```suggestion
                  key: validate-responses
                  name: management-api
```

harshad16 (changes requested):
Thanks for the addition.
Let a minor change requested based on the reference.

management-api/base/deploymentconfig.yaml
fridex: Nice catch, thanks! :+1:

harshad16 (commented):
/approve
This can go in .
Thanks for the addition.

PR #917, opened 2021-02-15
Introduce kafka env var update all across the envs

harshad16:
Introduce kafka env var update all across the envs

Fixes: https://github.com/thoth-station/thoth-application/issues/795


pacospace (commented):
One question only, otherwise looks good to me:

do  we want to remove env variable from stage for pending workflows limit?

```
            - name: ARGO_PENDING_WORKFLOW_LIMIT
              value: "2000"
```

harshad16:
> One question only, otherwise looks good to me:
> 
> do we want to remove env variable from stage for pending workflows limit?

yes, it was used for debugging purpose, it has served its need. now its not needed.

pacospace (approved):
Thanks! 💯

PR #925, opened 2021-02-16
Enable integration tests

fridex:

- [x] No

### DIFF
--- a/integration-tests/overlays/ocp4-stage/cronjob.yaml
+++ b/integration-tests/overlays/ocp4-stage/cronjob.yaml
@@ -7,4 +7,4 @@ metadata:
     app: thoth
     component: integration-tests
 spec:
-  suspend: true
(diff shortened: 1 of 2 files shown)


integration-tests/base/cronjob.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     component: integration-tests
 spec:
   schedule: "@daily"
-  suspend: true
+  suspend: false
```
harshad16, line 11: @fridex  can we let the base be suspended and turn on explicitly in each environment.
in this way, we would have flexibility in turning on or off , without concerning of base .

fridex: Makes sense :+1: Adjusted in ad5e104

harshad16 (approved):
/approve
Thanks for the chnages.

PR #934, opened 2021-02-18
clean stage overlays which corresponds to ocp cluster(no longer in use)

harshad16:
clean stage overlays which correspond to ocp cluster(no longer in use)

## Description

we have migrated completely to ocp4 cluster and have been using it for a while.
Based on polls, thoth-station team agreed we can remove these overlays from the master branch.


fridex (approved):
awesome :+1:

PR #935, opened 2021-02-18
Adjust command for integration-tests in Prow

fridex:

Fixes: https://github.com/thoth-station/integration-tests/issues/125

- [x] No

### DIFF
--- a/prow/overlays/cnv-prod/config.yaml
+++ b/prow/overlays/cnv-prod/config.yaml
@@ -305,10 +305,21 @@ periodics:
         - image: quay.io/thoth-station/integration-tests:v0.5.0
           imagePullPolicy: Always
           command:
-            - ./e2e-test.sh
+            - python3
+            - app.py
           env:
-            - name: E2E_TEST_API
+            - name: THOTH_DEPLOYMENT_NAME
+              value: cnv-prod
+            - name: MAIL_REPORT
+              value: "0"
+            - name: THOTH_USER_API_HOST
               value: api.moc.thoth-station.ninja
+            - name: THOTH_MANAGEMENT_API_HOST
+              value: management.moc.thoth-station.ninja
+            - name: THOTH_AMUN_API_HOST
+              value: amun.moc.thoth-station.ninja
+            - name: THOTH_MANAGEMENT_API_SECRET
+              value: ""
     annotations:
       description: Periodic e2e tests against MOC deployment
 


prow/overlays/cnv-prod/config.yaml
```diff
@@ -305,10 +305,21 @@ periodics:
...
+            - name: MAIL_REPORT
+              value: "0"
+            - name: THOTH_USER_API_HOST
               value: api.moc.thoth-station.ninja
+            - name: THOTH_MANAGEMENT_API_HOST
+              value: management.moc.thoth-station.ninja
+            - name: THOTH_AMUN_API_HOST
+              value: amun.moc.thoth-station.ninja
+            - name: THOTH_MANAGEMENT_API_SECRET
+              value: ""
```
fridex, line 322: Do we want to have this setup?

harshad16: oh yes we do have set these up. 
same as the stage as of now :see_no_evil:

harshad16: we can have these value set by configmap and secrets. As this component is not in thoth namespaces, so we can have these setup in prow namespace too explicitly.

fridex: Do we want to keep running these tests in Prow? I wonder if we could rather deploy integration-tests in a namespace running internally and point it to the prod deployment. It is a manner of a new overlay in integration-tests with adjusted setup. We will not need to maintain integration-tests in prow and we reuse the same logic we already have for test and stage environments (without a need to provide secrets here, have periodic e-mail reports). Does Prow give us any advantage here?

harshad16: I m in favor of running a cronjob, which sends out emails. It has an advantage added to it.
prow basically is just executing the cronjob, so I don't see any added advantage.

goern: prow advantage I see is, that it keeps all the logs around, and that we declare all the things wrt prod in one place.

prow/overlays/cnv-prod/config.yaml
```diff
@@ -305,10 +305,21 @@ periodics:
         - image: quay.io/thoth-station/integration-tests:v0.5.0
           imagePullPolicy: Always
           command:
-            - ./e2e-test.sh
+            - python3
```
goern, line 308: who is going to install the dependencies?

goern: see https://github.com/thoth-station/integration-tests/pull/136

fridex: > who is going to install the dependencies?

They should be already present in the container as part of the s2i build process.

goern: ja, .makes sense, thx!

PR #947, opened 2021-02-22
Do not run integration tests requiring middletier namespace

fridex:

With https://github.com/thoth-station/integration-tests/pull/137/ we are able to selectively run/disable selected tests based on tags. Let's do not run integration tests that require middletier resources in stage as stage is running ingestion in thoth-middletier-stage so tests fail. 

- [x] No

### DIFF
--- a/integration-tests/overlays/ocp4-stage/configmap.yaml
+++ b/integration-tests/overlays/ocp4-stage/configmap.yaml
@@ -8,3 +8,4 @@ data:
   user-api: "stage.thoth-station.ninja"
   management-api: "management.stage.thoth-station.ninja"
   amun-api: "amun.stage.thoth-station.ninja"
+  tags: "~@seizes_middletier_namespace"
--- a/integration-tests/base/configmap.yaml
+++ b/integration-tests/base/configmap.yaml
@@ -6,3 +6,4 @@ metadata:
 data:
   mail-report: "1"
   generate-report: "1"
+  tags: ""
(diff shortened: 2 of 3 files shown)


integration-tests/overlays/ocp4-stage/configmap.yaml
```diff
@@ -8,3 +8,4 @@ data:
   user-api: "stage.thoth-station.ninja"
   management-api: "management.stage.thoth-station.ninja"
   amun-api: "amun.stage.thoth-station.ninja"
+  tags: "~@seizes_middletier_namespace"
```
fridex, line 11: Note that even v2 tags are documented in the official docs:

https://behave.readthedocs.io/en/latest/tag_expressions.html#tag-expressions-v2

They do not work with [the latest stable release](https://stackoverflow.com/questions/62026928/python-behave-tags-or-logic-does-not-appear-to-be-working). See `--tags-help`:

```
Scenarios inherit tags declared on the Feature level. The simplest
TAG_EXPRESSION is simply a tag::

    --tags @dev

You may even leave off the "@" - behave doesn't mind.

When a tag in a tag expression starts with a ~, this represents boolean NOT::

    --tags ~@dev

A tag expression can have several tags separated by a comma, which represents
logical OR::

    --tags @dev,@wip

The --tags option can be specified several times, and this represents logical
AND, for instance this represents the boolean expression
"(@foo or not @bar) and @zap"::

    --tags @foo,~@bar --tags @zap.

Beware that if you want to use several negative tags to exclude several tags
you have to use logical AND::

    --tags ~@fixme --tags ~@buggy.
```

goern:
there is some conflict...

/assign @fridex

fridex:
> there is some conflict...
> 
> /assign @fridex

Thanks, rebased. :+1: 

/unassign @fridex

harshad16 (approved):
lgtm, thanks for this addition :100:

PR #958, opened 2021-02-24
Bump user-api to v0.19.2 in stage environment

fridex:

Related: thoth-station/user-api#1274

- [x] No


fridex:
Will replace with v0.19.3 that addresses pipfile parsing issue.

PR #968, opened 2021-02-24
Revert "Run integration tests that use backend namespace"

fridex:
Reverts thoth-station/thoth-application#967

We can revert this after Francesco's demo.


fridex:
/unhold
/assign @harshad16

I think we can proceed with this one.

harshad16 (commented):
/approve
thanks @fridex  :100:

PR #970, opened 2021-02-25
Adjust restart policy for integration tests

fridex:

As integration tests return non-zero exit code if any scenario fails, the job keeps restarting. We can run integration tests just once.

- [x] No


harshad16 (approved):
/lgtm 
valid value, can go in :+1:

Issue #991, opened 2021-02-25
Liveness probe fails on user-api when integration-tests are run


fridex:
**Describe the bug**

```
Readiness probe failed: Get "http://172.23.8.230:8080/readiness": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

**To Reproduce**
1. Run integration tests
2. See 503 being reported by user-api at the end of each integration-tests report
3. Check cluster events

**Expected behavior**

user-api should not be restarted

PR #992, opened 2021-02-25
Adjust web concurrency for user-api

fridex:

Fixes: https://github.com/thoth-station/thoth-application/issues/991

We boot 6 workers on 1 CPU which does not sound correct.

- [x] No


harshad16 (commented):
/lgtm
/approve

Issue #1000, opened 2021-03-01
🔥 FAILED syncing integration-tests components to TEST


fridex:
```
CronJob.batch "integration-tests" is invalid: spec.jobTemplate.spec.template.spec.restartPolicy: Unsupported value: "Always": supported values: "OnFailure", "Never"
```

PR #1002, opened 2021-03-01
state backofflimit inside jobtemplate for cronjob

harshad16:
state backofflimit inside jobtemplate for cronjob

Related-to: #1000

### DIFF
--- a/integration-tests/base/cronjob.yaml
+++ b/integration-tests/base/cronjob.yaml
@@ -15,13 +15,13 @@ spec:
   startingDeadlineSeconds: null
   jobTemplate:
     spec:
+      backoffLimit: 0
       template:
         metadata:
           labels:
             app: thoth
             component: integration-tests
         spec:
-          backoffLimit: 0
           restartPolicy: Never
           containers:
             - image: "integration-tests:latest"


integration-tests/base/cronjob.yaml
```diff
@@ -15,13 +15,13 @@ spec:
   startingDeadlineSeconds: null
   jobTemplate:
     spec:
+      backoffLimit: 0
```
harshad16, line 18: https://docs.openshift.com/container-platform/4.1/nodes/jobs/nodes-nodes-jobs.html#nodes-nodes-jobs-creating-cron_nodes-nodes-jobs

it seems like backofflimit is a parameter of job , so have moved it to jobtemplate spec.
@fridex  let me know if this is not required , we can remove it as well.

fridex (approved):
Nice catch! sorry for this bug :bug:

integration-tests/base/cronjob.yaml
fridex: @harshad16 I think this is a proper fix for integration-tests being run muliple times as the job gets restarted until backoff limit is reached

PR #1003, opened 2021-03-01
Include build-analysis template with package-extract component

harshad16:
Include build-analysis template with package-extract component

As build-analysis templates require package-extract imagestream and workflow-templates to be available, it would be convenient from the maintenance point to have them all in one component.


fridex (approved):
lgtm, thanks! :+1:

PR #1049, opened 2021-03-11
increase mem resource for the postgresql deployment

harshad16:
increase mem resource for the postgresql deployment


fridex (approved):
thanks :ok:

PR #1056, opened 2021-03-12
increase cpu for the postgresql in stage

harshad16:
increase cpu for the postgresql in stage

## Description

with this we would still have lot of free resources in the namespace thoth-graph-stage


fridex (approved):
thanks! :+1:

PR #1080, opened 2021-03-18
Fix links in graph-backup job

fridex:

- [x] No

Thanks to @harshad16 for spotting this.


harshad16 (approved):
Thanks

PR #1105, opened 2021-03-24
Run prescription sync command

fridex:

- [x] No

## Description

Otherwise only help message is printed as no command is supplied.


harshad16 (approved):
lgtm :tada:

PR #1115, opened 2021-03-26
Mi scheduler/reduce parallelism

xtuchyna:
## Related Issues and Dependencies
Two workflows are not operational. Reducing to only one :'(

## This introduces a breaking change

- [ ] Yes
- [x] No

## Test or Stage Environment

- [ ] Pull Requests added content to STAGE environment
- [ ] Pull Requests to `AICoE/aicoe-cd` has been opened:

PR #1140, opened 2021-04-02
Fix parallel limit to use semaphore instead of parallelism

xtuchyna:
Related to #1115 

- [ ] Yes
- [X] No

## Test or Stage Environment

## Description
The `parallelism` specification was indeed incorrect, and the right way to reduce number of parallel workflows that can run is to use `semaphore`


harshad16 (approved):
lgtm

PR #1162, opened 2021-04-12
set prow ssh cerds for private repo support

harshad16:
set prow ssh cerds for private repo support

Related-to: https://github.com/aicoe-aiops/vulnerability-analysis/pull/40

## Description

This would enable prow to use the ssh method when necessary, like serving the private repos.


tumido (commented):
/lgtm

PR #1163, opened 2021-04-12
Add missing User API token

fridex:

- [x] No

To enable authenticated workflow, we need to provide a token to User API. See:

https://github.com/thoth-station/user-api/blob/20ac6db4154e5480970727ff831d0272bfbd5d1c/thoth/user_api/configuration.py#L37

### DIFF
--- a/user-api/base/deploymentconfig.yaml
+++ b/user-api/base/deploymentconfig.yaml
@@ -38,6 +38,11 @@ spec:
               value: "khemenu.thoth-station.ninja"
             - name: "THOTH_SENTRY_IGNORE_EXCEPTION"
               value: "thoth.user_api.exceptions.NotFoundException,thoth.storages.exceptions.NotFoundError"
+            - name: THOTH_USER_API_TOKEN
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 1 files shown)


user-api/base/deploymentconfig.yaml
```diff
@@ -38,6 +38,11 @@ spec:
               value: "khemenu.thoth-station.ninja"
             - name: "THOTH_SENTRY_IGNORE_EXCEPTION"
               value: "thoth.user_api.exceptions.NotFoundException,thoth.storages.exceptions.NotFoundError"
+            - name: THOTH_USER_API_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  key: token
+                  name: thoth
```
fridex, line 45: This does not exist. Not sure if it should be mapped to another key. CC @harshad16

harshad16: please use the key `thoth-secret` for this.

fridex: Adjusted, PTAL

harshad16 (approved):
lgtm , thank you

PR #1190, opened 2021-04-21
add Kebechet slug to openshift templates (default null)

KPostOffice:

Fixes: https://github.com/thoth-station/investigator/issues/496

I added KEBECHET_SLUG to the OpenShift template to keep the above error from occuring. It is not currently being used in `prod` or `stage` but I added it to the base as they have a more recent version of `thoth-common` which attempts to set the template parameter.


harshad16 (approved):
Thanks for the fix :rocket:

PR #1201, opened 2021-04-26
Run kebechet-run-results only if authentication was done

fridex:

When running an advise workflow without authentication being set, the kebechet-run-results task should not be executed. Even though the task has a dependency on trigger-integration, argo evaluates `when` conditional for this task which results to an error. The workflow should not be marked as failed if the advise task succeeds but the kebechet-run-results conditional is evaluated as false.

The screenshot shows failed advise task, but if the advise succeeds, the kebechet-run-results fails with the same error marking the whole workflow as failed:

Screenshot_2021-04-26_13-03-15


pacospace (commented):
Thanks! 💯 lgtm

harshad16 (approved):
Thanks :100:

Issue #1202, opened 2021-04-26
Reduce the ttl for the workflow in thoth-backend-prod


harshad16:
**Describe the bug**
The workflow pods are staying on the namespace thoth-backend-prod, and hold the cpus 

![](https://files.slack.com/files-pri/T01SKMNPRUG-F020R0QPUSU/image.png)

we should reduce the TTL for the pods, to reduce the number of CPUs we hold.

fridex:
> we should reduce the TTL for the pods, to reduce the number of CPUs we hold.

I think the TTL strategy cleans pods. Once a pod finishes, it should not occupy CPU. We keep pods for logs that are available to users.

fridex:
BTW we should use logs that can be stored on ceph by argo using `archiveLogs: true`. That way we can provide logs to users for longer period of time and garbage collect pods sooner.

harshad16:
I completely agree that the CPU is not occupied after the pods completion. However it seems from the cluster , the resource pressure can be observed, intuition is request resource is still considered by the cluster (not really sure on this part).
https://github.com/operate-first/support/issues/223
I have switched on the archivelogs for us and reduce the ttl. Also we would have logs from loki stagging (as that is set as well).

PR #1219, opened 2021-05-03
Reduce the time to live 15 mins instead

harshad16:
Reduce the time to live 15 mins instead

Fixes: https://github.com/thoth-station/thoth-application/issues/1202

## Description

set archiving log true, for longer log retention without the issue of pod logs.
Reduce the time to live 15 mins instead


fridex (approved):
LGTM, let's see how we can utilize archived logs to improve user experience 👍🏻

PR #1223, opened 2021-05-03
turn the utility images into imagestreamtags

goern:

## Description

I turned the decorator utility images into imagestreamtags, so that we can save some pull-time on each job execution

/sig devops
/kind feature


harshad16 (approved):
thanks :+1:

PR #1355, opened 2021-05-19
set name of service port as metrics for the components

harshad16:
set name of service port as metrics for the components


fridex (approved):
💯 👍🏻

PR #1360, opened 2021-05-20
Set strict limit to test solver OOM error handling

fridex:

Related: https://github.com/thoth-station/solver/pull/704

## Description

Decreased to test if error handling implemented in https://github.com/thoth-station/solver/pull/704 will be done correctly.

### DIFF
--- a/solver/base/argo-workflows/solver-template.yaml
+++ b/solver/base/argo-workflows/solver-template.yaml
@@ -110,7 +110,7 @@ spec:
         resources:
           limits:
             cpu: 1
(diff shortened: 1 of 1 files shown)


fridex:
/hold

... until thoth-station/solver#704 is merged and new solver in test environment.

fridex:
I tested this locally. If we get OOM, solver should fail and no syncs should be done.

CC @pacospace we could see decreasing number of solver workflows - that will signalize that we do not have memory requirements set correctly in the deployment

/close

PR #1510, opened 2021-06-01
add peribolos for thoth-station

goern:
see https://github.com/kubernetes/test-infra/tree/master/prow/cmd/peribolos

### DIFF
--- a/prow/overlays/cnv-prod/peribolos.yaml
+++ b/prow/overlays/cnv-prod/peribolos.yaml
@@ -1,0 +1,42 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: peribolos
+spec:
+  backoffLimit: 2
+  template:
+    metadata:
+      name: peribolos
+    spec:
+      containers:
+        - name: peribolos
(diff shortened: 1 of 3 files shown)


tumido (commented):
/lgtm

prow/overlays/cnv-prod/peribolos.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: batch/v1
+kind: Job
```
tumido, line 2: @goern shouldn't this be a cron job? Do we plan to get it synced on config change? If yes and it would react on config update, we can use peribolos to create new repos/invite to org via PR and on demand.

goern: I thought of it as a Job so that we have more control over it during experiment/testing

goern: I think it should be a prow job that is run after merge to master (in this case) whenever the `org_*` file changes

tumido (commented):
/lgtm

goern:
I'll merge this for testing...

/approve

Issue #1414, opened 2021-05-24
sprint production release 2021.06.07


harshad16:
Hello, Thoth-station!

This Issue would be used for the current sprint cycle production release.
By the end of the sprint cycle, we will consolidate the information of thoth-station components features upgrade and fixes in this issue.

fridex:
Implemented handling of packages - selecting which packages should be analyzed by Thoth by maintaining solver rules: https://github.com/thoth-station/thoth-application/issues/1403

fridex:
Using PyPA's [advisory-db](https://github.com/pypa/advisory-db) as a source of security vulnerabilities.

cve-update-job: v0.3.0
adviser: v0.31.0

harshad16:
With these updates, we have completed the release of 2021.05.24

**Features** 

* Implemented handling of packages - selecting which packages should be analyzed by Thoth by maintaining solver rules: https://github.com/thoth-station/thoth-application/issues/1403

management-api: >0.17.0
adviser: >=0.31.0

* Using PyPA's [advisory-db](https://github.com/pypa/advisory-db) as a source of security vulnerabilities: https://github.com/thoth-station/cve-update-job/issues/404

cve-update-job: >=0.3.0
adviser: >=0.31.0

As this change required database schema changes, all the components that talk to the database need thoth-storages>=0.44.0 (updates propagated, new releases pending).

**Component Updates** 
- [x] adviser: v0.31.0
  - Keep CVE advisory in the justification output (https://github.com/thoth-station/adviser/pull/1882)
  - Adjust for CVEs consumed from PyPA (https://github.com/thoth-station/adviser/pull/1876)
  - Add a sieve implementing filtering based on solver rules (https://github.com/thoth-station/adviser/pull/1877)
  - Document solver rules (https://github.com/thoth-station/adviser/pull/1878)
  - Add video demonstrating constaints.txt use (https://github.com/thoth-station/adviser/pull/1870)
  - Handle error when all the direct deps are sieved (https://github.com/thoth-station/adviser/pull/1867)
  - Make boot pipeline unit responsible for providing stack info (https://github.com/thoth-station/adviser/pull/1863)
  - Report all unresolved dependencies (https://github.com/thoth-station/adviser/pull/1864)

- [x] management-api: v0.17.2
  - Delete solver rules endpoint returns a single rule delete (https://github.com/thoth-station/management-api/pull/770)
[...]

PR #1518, opened 2021-06-01
Use solver v1.9.0 in prod environment

fridex:
## Description

Older thoth-solver releases have a bug when dependencies are not handled properly on OOM.


harshad16 (approved):
lgtm,
thanks for the update
Related-to: #1414

PR #1544, opened 2021-06-03
add 'thamos_token' for user-api authentication

KPostOffice:

When we run Kebechet we need the whole authenticated workflow to run


fridex (approved):
lgtm 👍🏻

Issue #1516, opened 2021-06-01
Extend purge-job logic to handle documents placed on Ceph


fridex:
**Is your feature request related to a problem? Please describe.**

As Thoth administrator, I would like to make sure purge-job logic takes care of old documents stored on Ceph. This way we will keep only relevant JSON data present on Ceph under corresponding key (e.g. sync-job will not pick them up and sync them again once triggered).

**Describe the solution you'd like**

- [ ] use thoth-storages to query for solver document ids that are about to be deleted during purge
- [ ] move solver documents to another location (e.g. store them under another prefix)
   - [ ] handle adviser documents
   - [ ] handle solver documents
   - [ ] handle container image analyses documents
- [ ] remove entries from the database once all the relevant documents have been moved to another location
- [x] keep "to be depracated" solvers to the configmap
- [x] adviser notifies about soon to be depricated solvers
- [ ] update solver procedure document and state steps that need to be done each time solvers are removed/added - https://github.com/thoth-station/thoth-application/blob/master/docs/thoth_s2i_procedure.md

**Describe alternatives you've considered**

We could also delete the documents, but they can be a source of interesting data later (or have them available for experiments).

PR #1631, opened 2021-06-08
Pass configured solvers in a deployment to adviser

fridex:

Related: https://github.com/thoth-station/thoth-application/issues/1516

## Description

As discussed on the call, we can pass solvers to the adviser and let the adviser decide whether the given runtime environment has solvers enabled in the configmap. If they will not be present there, adviser can warn about the fact in a separate pipeline unit. This might be better to maintain than having two configuration entries for solvers enabled & solvers deprecated.


harshad16 (approved):
lgtm, thanks :100:

PR #1641, opened 2021-06-09
Add increased timeout seconds

xtuchyna:
None

- [X] Deployment for Test

## Description
`mi` workflows were suspended because the one that was running waited too long for GH API limit to be reset (which is once per hour).

This PR should extend the timeout time up to one hour

### DIFF
--- a/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
+++ b/mi-scheduler/base/argo-workflows/mi-workflow-template.yaml
@@ -8,6 +8,7 @@ metadata:
     template: mi
 
 spec:
+  activeDeadlineSeconds: 3600
   serviceAccountName: argo
   entrypoint: mi-workflow
   templates:
@@ -94,5 +95,3 @@ spec:
             value: "."
           - name: THOTH_ADJUST_LOGGING
             value: "{{inputs.parameters.THOTH_ADJUST_LOGGING}}"
-        livenessProbe:
-          timeoutSeconds: 3600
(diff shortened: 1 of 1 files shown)


xtuchyna:
One thing I am not sure about is the documentation on these specifications:
```
periodSeconds: The delay, in seconds, between performing probes. The default is 10. This value must be greater than timeoutSeconds.

timeoutSeconds: The number of seconds of inactivity after which the probe times out and the container is assumed to have failed. The default is 1. This value must be lower than periodSeconds.
```
However, we do not follow this rule e.g. here:
https://github.com/thoth-station/thoth-application/blob/340794bd12631c9e5d8fd75cf0ffc4827956cee5/adviser/base/argo-workflows/advise.yaml#L241

Does this not apply for workflows?

goern:
@tumido wdyt?

xtuchyna:
so this was the log of the failed pod
image

xtuchyna:
so the `activeDeadlineSeconds` is probably the right one to use here
image

/cc @goern 
/cc @harshad16

harshad16 (commented):
/lgtm
we could try this :+1:

tumido:
Sorry for being late here, yeah, I think that's fine.. in Aiops Argo we set this to 8 hours. I think that greatly depends on the expected runtime duration of the workflow. We expect it to finish sooner than 8 hours, if it should take longer, we default to failure.

Please know that `activeDeadlineSeconds` will kill any running pod if it's running longer than this timeout. Don't confuse it with healthchecks (like `timeoutSeconds`), because it doesn't care about pod health or heartbeat or anything, it just kills it..

https://argoproj.github.io/argo-workflows/fields/#fields_3

xtuchyna:
@harshad16 could we /approve this, then?

harshad16 (approved):
thanks :100:

PR #1665, opened 2021-06-11
Block security indicator step to resolve stacks without si

fridex:

Related: https://github.com/thoth-station/adviser/issues/1911


fridex:
I will block the security indicator step. I would like to test the behavior of resolving software stacks without CVE. Let's disable security indicator step as we do not have security indicator data aggregated.

CC @KPostOffice 

/approve

goern:
@KPostOffice could you please create an issue that outlines what needs to be don't to include SI in advise

Issue #1649, opened 2021-06-10
deploy triage-party


goern:
**Is your feature request related to a problem? Please describe.**
As a Thoth-Station Devs, I want to run a triage party, so that issues/pr get triaged daily, weekly, quarterly, ...

**Describe the solution you'd like**

 - [x] merge #1650 
 - [x] create project on zero-prod @harshad16  (we don't need it, we will use thoth-bot-prods)
 - [x] create DNS RR `triage-party.thoth-station.ninja` @goern 
 - [x] create secret and tls route in `overlays/zero-prod` @harshad16 
-  [ ] Manually deploying the app @Gregory-Pereira 
 - [ ] create ArgoCD app @Gregory-Pereira 
 - [ ] add missing item to this list

**Additional context**
This is a small web application that will help us triage issue and PR more easily

/kind feature
/priority importnat-soon
/assign @harshad16 
/kind demo
/assign @goern

harshad16:
This image doesn't exists: `quay.io/goern/triage-party` , how was the image created @goern ?

goern:
image exists, repo was private, I reconfigured to public. pls chk.

Gregory-Pereira:
I am trying to deploy triage party as an application on the thoth-bots-prod namespace on the Zero cluster. After deploying I am getting this error: 
"ERROR: don't know how to run your application.
Please set either APP_MODULE, APP_FILE or APP_SCRIPT environment variables, or create a file 'app.py' to launch your application."

goern:
It might be that the trick it: this a Go app. the error you see is mostly associated with Python apps. How did you deploy it?

Gregory-Pereira:
First I tried deploying it regularly as a python app which didn't work of course. Then I tried using the openshift/golang:latest operator from the CLI with and without the command google uses to start their triage party: 
```
go run cmd/server/main.go \
  --github-token-file=$GITHUB_TOKEN \
  --config config/examples/kubernetes.yaml \
  --repos kubernetes/sig-release
```
One thing I was confused about though is that I know I ran it as a golang application from the CLI but in the web console I only have the installed operators view, instead of the ability to add them (I think its related to my roles / permissions), so not sure how it would have deployed as such from the CLI.

harshad16:
while i applied the manifest present in the triage-party directory.
the errors I m observing are:
```
W0615 17:00:09.615319       1 main.go:90] --config and CONFIG_PATH were empty, falling back to /app/config/config.yaml
[...]

PR #1667, opened 2021-06-11
deploy triage-party #1649

Gregory-Pereira:

Deployment failed because of incorrect storage type

- [ x ] Deployment for Test and Stage `AICoE/aicoe-cd` and Prod `operate-first/argocd-apps`.

## Description

When manually deployed it failed because the storage instance was Ceph instead of noobaa.


harshad16 (approved):
lgtm, thanks :100:

Issue #1686, opened 2021-06-14
Provision new PVC for prescriptions


fridex:
**Is your feature request related to a problem? Please describe.**

As of now we use Argo to obtain `prescription.yaml` from Ceph:

https://github.com/thoth-station/thoth-application/blob/b305c19f230b1a948d854371ed910ae062d0a63e/adviser/base/argo-workflows/advise.yaml#L87-L102

This is something we'll need to change with https://github.com/thoth-station/adviser/issues/1908. Once the change lands, prescription-sync-job will store content of the prescriptions Git repository to a PV which would get mounted into each adviser container.

- [ ] prescription-sync-job will have the PV mounted in `rw` mode
- [ ] each adviser will have PV mounted in a `r` mode

**Describe the solution you'd like**

Provision a persistent volume in all the deployments (test/stage/prod).

fridex:
/assign @harshad16 
/assign @goern 

Is this something we can accomplish in our deployments?

fridex:
Also, if the prescription-sync-job will be run same time as adviser, could data collide?

harshad16:
I will read up on this, and checking on the requirements from cluster side.

fridex:
> I will read up on this, and checking on the requirements from cluster side.

Thanks! 👍🏻 Just to sum up from the call, maybe @tumido can help on this front directly from top of his head:

* we need persistent volume in Thoth deployment
* the persistent volume will be mounted once by a cron job that will periodically place data there - the volume will be mounted in `write` mode only by one pod at a time
* the persistent volume will be mounted multiple times by pods that will use data placed there

The only mode that satisfies this is [ReadWriteMany](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) as our requirements are subset of ReadWriteMany. Not sure if there is such PV on op1st.

fridex:
Just a note - might be better to just build a container with data and share data with the main container in deployment.

fridex:
> Just a note - might be better to just build a container with data and share data with the main container in deployment.

I think we can close this with this solution. Thanks and sorry for tags :)

/close

PR #1693, opened 2021-06-15
Decommission prescription-sync-job

fridex:

It will be replaced with a prescription container - see https://github.com/thoth-station/thoth-application/issues/1686#issuecomment-861460845


harshad16 (approved):
/unhold
/approve
thanks :100:

PR #1695, opened 2021-06-15
Make prescriptions available to the adviser container

fridex:
## Related Issues and Dependencies

Related: https://github.com/thoth-station/prescriptions/pull/24
Depends-On: https://github.com/thoth-station/thoth-application/pull/1694
Depends-On: https://github.com/thoth-station/thoth-application/pull/1701
Related: https://github.com/thoth-station/adviser/issues/1908

PR #1701, opened 2021-06-16
Disable prescriptions before adding prescriptions with dir structure

fridex:

Related: https://github.com/thoth-station/thoth-application/pull/1695/

As new prescriptions with directory structure are not backwards compatible with old, single file, prescriptions, let's disable old ones. Once the transition is made, we can enable prescriptions with directory structure https://github.com/thoth-station/thoth-application/pull/1695/.


fridex:
CC @harshad16 this will need to go in before we add the "prescriptions" container and new adviser release PTAL

harshad16 (approved):
/lgtm
thanks :100:

PR #1745, opened 2021-06-22
fix of the volume placement in the workflow

harshad16:
fix of the volume placement in the workflow

Related-to: #1743 

## Description

This is tested :medal_sports:


fridex:
@harshad16 I wonder why the old solution worked in the test environment. Anyway, thanks 👍🏻

harshad16:
> I wonder why the old solution worked in the test environment. Anyway, thanks 👍🏻

i wonder the same, need to have a closer look at test environment.

Issue #325, opened 2020-08-26
Archiving and Retrieving Workflow Logs


KPostOffice:
We want our workflows to be automatically deleted after they are completed whether they succeed or fail for the most part. Leaving them to live in the workflow controller and leaving the pods on OpenShift causes unnecessary resource usage.  However, having pod logs is important for debugging. WorkflowController has a optional configuration for automatically storing logs by [setting up a bucket](https://raw.githubusercontent.com/argoproj/argo/master/docs/workflow-controller-configmap.yaml) and setting [archiveLogs to be true](https://github.com/argoproj/argo/blob/master/examples/archive-location.yaml) in workflows.  We should store logs and set up a way for retrieving these logs and also deleting them after a certain number of hours or days.

PR #1758, opened 2021-06-23
Decrease TTL to 300 for pods

fridex:

As we keep logs on Ceph, we can decrease TTL and cleanup pods sooner.

Related: https://github.com/thoth-station/thoth-application/issues/325


pacospace (approved):
lgtm!

PR #1782, opened 2021-06-30
feat: Use ACME operator for routes

tumido:
After some fighting for over a month with ACME controller in Operate First we finally made it operational. We now use it for ArgoCD and ODH Dashboard routes successfully https://github.com/operate-first/apps/pull/780

This PR makes Prow to use it as well, which means from now on there should be no need for manual cert renewals for prow as well!

/cc @harshad16


tumido:
bump @harshad16 wanna take a look at this?

harshad16:
> bump wanna take a look at this?

waiting for the Friday , so if it fails anything we have time to fix :sweat_smile: 
thanks for the pr, i will merge it on friday.

PR #1809, opened 2021-07-08
Add MI entities as PullRequest and Issue

xtuchyna:
## Related Issues and Dependencies
None

## Does this require new deployment ?

- [ ] Deployment for Test and Stage `AICoE/aicoe-cd` and Prod `operate-first/argocd-apps`.

## Description

All of the data aggregation requests are currently Issues and Pullrequests, therefore changing entities to be only these by default

PR #1810, opened 2021-07-08
Fix reference to be key instead of value

xtuchyna:
Related to https://github.com/thoth-station/thoth-application/pull/1809


pacospace (approved):

pacospace:
> /cc @pacospace

I thought kustomize build would catch that

harshad16:
> I thought kustomize build would catch that

kustomize can catch only structural issues, these values are not know to kustomize i think.

PR #1821, opened 2021-07-12
feat: Enable tide on meteor and require approved on common

tumido:
Seems like Tide was not enabled in AICoE/meteor repo https://github.com/AICoE/meteor/pull/39

This is fixing it. Also switches for meteor and common repo to requiring both `lgtm` and `approved` labels.


pacospace (commented):
/lgtm

PR #1828, opened 2021-07-13
:sparkles: do not include area/rel-eng issues from filter

goern:


harshad16 (approved):
like it :+1: 
/lgtm
/approve

PR #1891, opened 2021-08-18
Increase package-releases deadline to 4 hours

fridex:

It looks like we require more time to check all the packages that we monitor.


pacospace (approved):
Thanks! lgtm! 💯

Issue #1892, opened 2021-08-18
Run a solver task that has more resources allocated if the default solver task fails


fridex:
**Is your feature request related to a problem? Please describe.**

See https://github.com/thoth-station/support/issues/11, also related OOM kills of solver jobs that can be spotted from time to time in the cluster.

The goal here is to create a separate task, aside from the main solver task, that will rerun solver container with more resources allocated if the main solver task timeouts or is killed on OOM. The rest of the workflow can be reused (e.g. syncing data into the database).

**Describe the solution you'd like**

- [x] create a solver task inside already existing solver workflow
- [x] the newly created solver task will have more CPU and memory resources allocated so that it can solve packages that require more resources
- [x] run the newly created solver task as an alternative to the main solver task, only if the main solver task fails

PR #1894, opened 2021-08-18
Branch workflow with more resources if solver fails

fridex:

Fixes: https://github.com/thoth-station/thoth-application/issues/1892

## Description

Implement a new branch of the solver workflow that will retry solver with more resources when solver fails. The solver might fail due to not enough resources needed when compiling packages from the source.

Sadly Argo does not support parametrizing resources. Also, using anchors and YAML aliases just complicated things. If anyone has a better solution not to copy-paste parts of the workflow, suggestions are welcome.

### DIFF
--- a/solver/base/argo-workflows/solver-template.yaml
+++ b/solver/base/argo-workflows/solver-template.yaml
@@ -39,8 +39,6 @@ spec:
               value: "{{duration}}"
 
       resubmitPendingPods: true
-      retryStrategy:
-        limit: 2
       inputs:
         parameters:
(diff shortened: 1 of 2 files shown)


pacospace (commented):
lgtm! 💯

fridex:
> lgtm!

Tested in the test environment. As ingestion in the stage is stalled on this, let's give this a try also in the stage environment.

/approve

fridex:
Looks like this did the trick and we continue ingesting:

Screenshot_2021-08-19_10-57-32

PR #1898, opened 2021-08-19
Set deadline for solver containers

fridex:


solver/base/argo-workflows/solver-template.yaml
```diff
@@ -70,6 +70,7 @@ spec:
               secretKeySecret:
                 name: argo-artifact-repository-secrets
                 key: secretKey
+      activeDeadlineSeconds: 600
```
pacospace, line 73: From the data we collect not 100% of solvers finish in 600s, it s only 97%, and 99% below 900s.

fridex: Nice, adjusted to 900. 👍🏻

pacospace (approved):
Thanks!

PR #1964, opened 2021-09-15
add sesheta-srcops secrets

goern:
closes #1963 

Signed-off-by: Christoph Görn <goern@redhat.com>

PR #1965, opened 2021-09-15
create new workflow for repo initialization

KPostOffice:
 Depends on #1964 

## Description
I have added a new template for running advise and adding `.thoth.yaml` file to uninitialized repositories. This may not be the best place for this in `thoth-application` please let me know if you feel it is better of living somewhere else.

### DIFF
--- a/thoth-repo-init/base/argo-workflows/thoth-repo-init.yaml
+++ b/thoth-repo-init/base/argo-workflows/thoth-repo-init.yaml
@@ -1,0 +1,58 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: thoth-repo-init
+spec:
+  templates:
+    - name: thoth-repo-init
+      resubmitPendingPods: true
+      inputs:
+        parameters:
+          - name: "project-url"
+      container:
+        image: workflow-helpers:latest
+        volumeMounts:
+          - name: ssh-config
+            mountPath: /opt/app-root/src/.ssh
+            readOnly: true
+        env:
+          - name: THOTH_DEMO_PROJECT_URL
+            value: "{{inputs.parameters.project-url}}"
+          - name: THOTH_WORKFLOW_TASK
+            value: "thoth_repository_initialization"
+          - name: GIT_SSH_COMMAND
+            # Needed to keep weak host verification for git clone.
+            value: "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
+          - name: THOTH_DEMO_FORK_NAMESPACE
+            valueFrom:
+              configMapKeyRef:
+                name: thoth
+                key: fork-namespace
+          - name: GITHUB_TOKEN
+            valueFrom:
+              secretKeyRef:
+                key: github-access-token
+                name: thoth-srcops
+          - name: SENTRY_DSN
+            valueFrom:
+              secretKeyRef:
+                key: sentry-dsn
+                name: thoth
+          - name: THOTH_DEPLOYMENT_NAME
+            valueFrom:
+              configMapKeyRef:
+                key: deployment-name
+                name: thoth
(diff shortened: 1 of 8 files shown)


harshad16 (commented):
lgtm
thanks

@pacospace  can PTAL before we merge this
thanks

adviser/base/argo-workflows/advise-demo.yaml
```diff
@@ -0,0 +1,53 @@
...
+          - name: ssh-config
+            mountPath: /home/user/.ssh
+            readOnly: true
+        env:
+          - name: THOTH_DEMO_PROJECT_URL
+            value: "{{inputs.parameters.project-url}}"
+          - name: THOTH_WORKFLOW_TASK
+            value: "thoth_advise_demo"
+          - name: THOTH_DEMO_FORK_NAMESPACE
+            value: "{{inputs.parameters.fork-namespace}}"  # CHNG: get value from config map or pass directly from Thoth common?
```
KPostOffice, line 26: @harshad16 I want to create a configmap entry for this value rather than making it a template parameter. Which configmap should I add it to?

harshad16: If it is dependent on cluster, it can go in the **configmap thoth** under **common** in each **overlays** https://github.com/thoth-station/thoth-application/tree/master/core/overlays
or else you can add it to the **base configmap thoth**

adviser/base/argo-workflows/advise-demo.yaml
```diff
@@ -0,0 +1,58 @@
...
+              configMapKeyRef:
+                name: thoth
+                key: logging-no-json
+        resources:
+          limits:
+            cpu: "0.25"
+            memory: "256Mi"
+          requests:
+            cpu: "0.1"
+            memory: "128Mi"
```
harshad16, line 58:
```suggestion
            cpu: "250m"
            memory: "256Mi"
          requests:
            cpu: "100m"
            memory: "128Mi"
```
please use this notation, instead of decimal, I m not sure that works.

harshad16 (changes requested):
lgtm
however, would we need this in the adviser or we should keep this separate
so can be deployed with other applications?

adviser/base/openshift-templates/advise-demo.yaml
```diff
@@ -0,0 +1,100 @@
...
+        - name: ssh-config
+          secret:
+            secretName: thoth-srcops
+            items:
+            - key: ssh-privatekey
+              path: id_rsa
+      templates:
+        - name: "thoth-advise-demo"
+          archiveLocation:
+            archiveLogs: true
```
fridex, line 90: Do we want to keep logs in this case?

adviser/base/openshift-templates/advise-demo.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: advise-demo
```
fridex, line 4: It might be a good idea to remove the `demo` part and substitute it with `repo-initialization` or something like that from the beginning.

thoth-repo-init/base/argo-workflows/thoth-repo-init.yaml
```diff
@@ -0,0 +1,58 @@
...
+  name: thoth-repo-init
+spec:
+  templates:
+    - name: thoth-repo-init
+      resubmitPendingPods: true
+      inputs:
+        parameters:
+          - name: "project-url"
+      container:
+        image: image-registry.openshift-image-registry.svc:5000/thoth-test-core/workflow-helpers:latest
```
harshad16, line 14:
```suggestion
        image: workflow-helpers:latest
```

we can use this, as we would be patching this with kustomize file.

harshad16 (approved):
lgtm
thanks :100:

PR #1979, opened 2021-09-22
Add Pep's key to sops

codificat:

Not a deployment, but it would require the existing secrets to be re-encrypted with the key `8D191B7544E9CCC3547B25A877E5A5B31D13A647`.

Maybe we can do this on an as-needed basis though

## Description

This adds my key fingerprint to the various `.sops.yaml` config files.


codificat:
@harshad16 FYI

PR #1995, opened 2021-10-01
Remove latest from cve-update

pacospace:
Screenshot from 2021-10-01 11-35-12

### DIFF
--- a/cve-update/base/cronjob.yaml
+++ b/cve-update/base/cronjob.yaml
@@ -23,7 +23,7 @@ spec:
         spec:
           restartPolicy: OnFailure
           containers:
-            - image: "cve-update-job:latest"
+            - image: "cve-update-job"
               name: cve-update
               env:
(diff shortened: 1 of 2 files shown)


codificat:
The problem here is that the ImageStream is not being looked up. The reference itself looks correct; it has not changed in months, and jobs were running correctly until yesterday.

I am not sure what changed in the cluster, but I believe it's not caused by the imagestream reference in the cronjob definition. A workaround could be to directly reference the image itself (a different change that the currently proposed) but hopefully we can fix the actual root cause instead, as this might be affecting other ImageStream lookups.

/hold
for now

harshad16:
The `latest` defines the openshift component to explicitly look for `latest` tag in the imagestream.
Not having explicitly mention of tag, also results in looking for latest tag in the imagestream

i don't see any issue in here.
closing this PR. 
please open an issue and lets discuss the solution there.

PR #2066, opened 2021-10-20
Turn off ingestion in the moc cluster

harshad16:
Turn off ingestion in the moc cluster

## Description

we are running ingestion is **Stage**
and Prod is also turned on.
Let's turn off this one.


harshad16:
> lgtm, one coment: I think also `package release job` needs to be turned off, because it sends messages for scheduling solvers, wdyt?

@pacospace  that one is already turned off : https://github.com/thoth-station/thoth-application/blob/f64294eab997c1df2b2d96baa293039b8a0d722a/package-releases/overlays/moc-prod/cronjob.yaml#L11

Issue #2029, opened 2021-10-11
Setup  promethus  and grafana  service for service metrics


harshad16:
Setup Prometheus and Grafana service for collecting the Services Metrics 

**Acceptance criteria:**
- [x] verify thoth advise are functioning.
- [x] verify API services are available.
- [x] verify integration tests are running.

Gregory-Pereira:
Metrics in balrog:
    Grafana:
        - [x] update the grafana data source from balrog cluster to smaug grafana: merged in [operate-first/apps pr 1271](https://github.com/operate-first/apps/pull/1271)

harshad16:
- [ ] Fix the datasource https://github.com/thoth-station/thoth-application/blob/b9e4626ef37d1af4c491dea332b7fd274840875d/grafana-dashboard/overlays/moc-prod/thoth-workflow-controllers-metrics.json#L32
- [ ] Remove job field from the grafana dashbaord: https://github.com/thoth-station/thoth-application/blob/b9e4626ef37d1af4c491dea332b7fd274840875d/grafana-dashboard/overlays/aws-prod/thoth-knowledge-graph-content-metrics.json#L170
- [ ] Prometheus Rules in the global-project. 
![promrules](https://user-images.githubusercontent.com/14028058/140398812-f18968d5-6df3-4c4c-8496-4f09d00adb3c.png)
https://github.com/operate-first/apps/blob/c8bf9cf83471f221bc1d8f0944b4a473dc622611/argocd/overlays/moc-infra/projects/global_project.yaml#L165

harshad16:
Another issues related to this: https://github.com/thoth-station/thoth-application/issues/1004

goern:
@Gregory-Pereira  @harshad16 is this something we can deliver/close in the next milestone?

harshad16:
This is complete and all the dashboards are in place.
for alarms there is another issue: https://github.com/thoth-station/thoth-application/issues/1914
closing this. Thanks for all the work.

PR #2108, opened 2021-10-31
Infra-prod service-monitor creates field label on balrog

Gregory-Pereira:

Addresses part of [issue 2029](https://github.com/thoth-station/thoth-application/issues/2029)
    - [_4/5_] create service monitor for each namespace corresponding components
Deploys this service monitor in the same way as [pr 2038](https://github.com/thoth-station/thoth-application/pull/2038/) but for balrog.

## Description

This PR adds enables our service label for metrics from thoth-infra-prod to have a new field label with the path/route of the corresponding service as the value through a service monitor deployed to balrog.


harshad16 (approved):
lgtm
thanks :100:

PR #2190, opened 2021-12-14
Run prescriptions-refresh part that computes Thoth community stats

fridex:
## Description

Let's run handler responsible for computing Thoth community stats and creating prescriptions out of them.


fridex:
/hold

... until we will have more community data available.

goern:
what is the status of this one?

Issue #2172, opened 2021-11-30
reduce request/limit of cpu/memory on aws-prod deployment


goern:
**Is your feature request related to a problem? Please describe.**
looking at https://grafana.operate-first.cloud/d/9iXbjpD7k/namespace-cpu-and-memory-overview?orgId=1&var-datasource=aws-balrog-monitoring&var-namespace=thoth-frontend-prod&var-cluster=emea%2Fbalrog we could go ahead and significantly reduce cpu/memory request/limit for the deployment of user-api on aws-prod

**Describe the solution you'd like**
reconfigure user-api and other components

please note in this issue what the current request/limit of components is, and to what new values we can reduce it. so that we see what the resource consumption effect of solving this issue is.

goern:
same seem to be true for https://grafana.operate-first.cloud/d/9iXbjpD7k/namespace-cpu-and-memory-overview?orgId=1&var-datasource=aws-balrog-monitoring&var-namespace=thoth-middletier-prod&var-cluster=emea%2Fbalrog

goern:
and https://grafana.operate-first.cloud/d/9iXbjpD7k/namespace-cpu-and-memory-overview?orgId=1&var-datasource=aws-balrog-monitoring&var-namespace=thoth-infra-prod&var-cluster=emea%2Fbalrog

goern:
and https://grafana.operate-first.cloud/d/9iXbjpD7k/namespace-cpu-and-memory-overview?orgId=1&var-datasource=aws-balrog-monitoring&var-namespace=opf-kafka&var-cluster=emea%2Fbalrog

goern:
any updates on this? do we consider it done?

PR #2215, opened 2021-12-20
Freeze the data ingestion on balrog prod cluster

harshad16:
Freeze the data ingestion on balrog prod cluster

Related-to: https://github.com/thoth-station/thoth-application/issues/2172

## Description

As discussed in the Tech Talk the data ingestion would be carried out on stage.
So freezing the ingestion in prod
along with making the db dump of prod weekly.

### DIFF
--- a/cve-update/overlays/aws-prod/cronjob.yaml
+++ b/cve-update/overlays/aws-prod/cronjob.yaml
@@ -8,7 +8,7 @@ metadata:
     component: cve-update
 spec:
   schedule: "@daily"
-  suspend: false
+  suspend: true
   successfulJobsHistoryLimit: 2
   failedJobsHistoryLimit: 4
   concurrencyPolicy: Forbid
--- a/graph-backup-job/overlays/aws-prod/cronjob.yaml
+++ b/graph-backup-job/overlays/aws-prod/cronjob.yaml
@@ -3,5 +3,9 @@ kind: CronJob
 apiVersion: batch/v1beta1
 metadata:
   name: graph-backup
+  labels:
+    app: thoth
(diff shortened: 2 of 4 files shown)


codificat:
Just a couple of quick questions/verifications, for my own education: 

> So freezing the ingestion in prod

So, all ingestion comes "only" from `cve-update`, `package-update`, and `graph-refresh` then, right?  Somehow I thought there was something else, but clearly I had a wrong assumption.

The same question in a different way: what would be the best documentation place to answer the above? Asking for a friend :innocent: 

> along with making the db dump of prod weekly.

and for more frequent backups we rely on stage (as the source of ingested data), correct?

Thanks

harshad16:
> Just a couple of quick questions/verifications, for my own education:
> 
> > So freezing the ingestion in prod
> 
> So, all ingestion comes "only" from `cve-update`, `package-update`, and `graph-refresh` then, right? Somehow I thought there was something else, but clearly I had a wrong assumption.
> 

Yes that is correct , only these along with package-release is responsible for ingestion. That is already suspended.

> The same question in a different way: what would be the best documentation place to answer the above? Asking for a friend innocent

This would be the documentation: 
https://thoth-station.ninja/docs/developers/adviser/architecture.html?highlight=architecture#frontend-namespace
https://github.com/thoth-station/adviser/blob/master/docs/source/architecture.rst#frontend-namespace

if we need info on the architecture.

> > along with making the db dump of prod weekly.
> 
> and for more frequent backups we rely on stage (as the source of ingested data), correct?

yes, stage dumps are daily.

codificat:
Many thanks.
/approve

PR #2223, opened 2021-12-21
Enable the transfer-issue prow plugin

codificat:

An example of a `/transfer` request (currently not working because the plugin is not enabled): https://github.com/thoth-station/kebechet/issues/942#issuecomment-998797918

## Description

This is to enable the [transfer-issue hook plugin](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/transfer-issue) in prow so that org members can ask prow to move issues across repos within the org


harshad16 (approved):
lgtm
thanks :100:

PR #2227, opened 2021-12-21
Increase number of seconds allowed package-extract to run

fridex:

Some container images take way too long to be analyzed and exceed limit configured.


harshad16 (approved):
lgtm
thanks :100:

PR #2228, opened 2021-12-22
Increase memory requirements for package-extract

fridex:
## Description

Some container image analyses fail on OOM. Increase allocated memory for package-extract.


pacospace (approved):
lgtm! 💯

PR #2247, opened 2022-01-10
Adjust links to Thoth search UI based on deployment

fridex:
## Description

Configure links to Thoth search UI based on deployment used.

### DIFF
--- a/adviser/overlays/moc-prod/configmap.yaml
+++ b/adviser/overlays/moc-prod/configmap.yaml
@@ -7,3 +7,4 @@ data:
   validate-prescription-schema: "0"
   validate-unit-configuration-schema: "0"
   blocked-units: "SecurityIndicatorStep,ThothS2IAbiCompatibilitySieve"
+  search-url: "https://thoth-station.ninja/search/{document_id}/summary"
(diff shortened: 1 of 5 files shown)


harshad16 (approved):
lgtm
thanks :100:

adviser/overlays/moc-prod/configmap.yaml
```diff
@@ -7,3 +7,4 @@ data:
   validate-prescription-schema: "0"
   validate-unit-configuration-schema: "0"
   blocked-units: "SecurityIndicatorStep,ThothS2IAbiCompatibilitySieve"
+  search-url: "https://thoth-station.ninja/search/{document_id}/summary"
```
fridex, line 10: @harshad16 I'm unsure how we want to treat these deployments. We have one Thoth Search UI for prod, so not sure if we want to point users to the same Thoth Search UI in `aws-prod` and `moc-prod`. WDYT?

CC @Gkrumbach07

harshad16: i think we should only point users for `aws-prod` as it would be along the khemenu.thoth-station.ninja 
however we can have multiple deployments in `moc-prod` and stage for our experiments.

PR #2267, opened 2022-01-18
Update triage-party configuration with more views

codificat:
This is to add the following additional boxes to triage-party:

 - inconsistencies: help+important, inconsistent triage
 - queue information: what's active, what's next

Also adding the helm-charts repo

### DIFF
--- a/triage-party/overlays/moc/config.yaml
+++ b/triage-party/overlays/moc/config.yaml
@@ -13,9 +13,11 @@ settings:
     - https://github.com/thoth-station/common
     - https://github.com/thoth-station/core
     - https://github.com/thoth-station/cve-update-job
+    - https://github.com/thoth-station/datasets
     - https://github.com/thoth-station/graph-backup-job
     - https://github.com/thoth-station/graph-refresh-job
     - https://github.com/thoth-station/graph-sync-job
+    - https://github.com/thoth-station/helm-charts
     - https://github.com/thoth-station/init-job
     - https://github.com/thoth-station/invectio
     - https://github.com/thoth-station/investigator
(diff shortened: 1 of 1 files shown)


triage-party/overlays/moc/config.yaml
```diff
@@ -547,6 +535,23 @@ rules:
       - label: "!.*bug"
       - label: ".*question"
 
+  up-next:
+    name: "critical/urgent issues not active yet (i.e. what's next)"
+    resolution: "Assign if needed, set lifecycle/active when work starts"
+    type: issue
+    filters:
+      - label: "!lifecycle/active"
+      - label: "priority/critical"
```
goern, line 544: this is critical-urgent

triage-party/overlays/moc/config.yaml
```diff
@@ -595,6 +600,39 @@ rules:
...
+      - label: "lifecycle/active"
+      - tag: "!assigned"
+
+  long-active:
+    name: "Lifecycle Active but not updated for long"
+    resolution: "Update the status"
+    type: issue
+    filters:
+      - label: "lifecycle/active"
+      - updated: +14d
```
goern, line 618: shall we make this 10d, so that it is not spanning two sprints (at least formally not)?

triage-party/overlays/moc/config.yaml
```diff
@@ -20,6 +20,7 @@ settings:
     - https://github.com/thoth-station/invectio
     - https://github.com/thoth-station/investigator
     - https://github.com/thoth-station/jupyterlab-requirements
+    - https://github.com/thoth-station/helm-charts
```
goern, line 23: good point, maybe check if we are missing more repos?!

goern (changes requested):

triage-party/overlays/moc/config.yaml
codificat: Right... I was being lazy, as this matches as a regular expression, but I'll do the needful

triage-party/overlays/moc/config.yaml
codificat: We certainly are: the list here has 50 repos, and there are 128 (public and not archived).

I will take a closer look at which ones to add... or should we add them all?

goern: I think not all are required... let's add the obvious and keep on reviewing

codificat: For now I am adding a few more that have active issues.

PR #2270, opened 2022-01-19
Enable pulp-pypi-sync-job in test environment

fridex:


harshad16 (commented):
lets keep this turned-off
and get a deployment on stage namespace for better flow
in this way we would be able to keep testing namespace of testing 
and stage for actually generating data.
wdyt?

fridex:
OK, sounds good and reasonable. It can be triggered manually from the command line as needed. 👍🏻 

/close

PR #2275, opened 2022-01-24
Remove prescriptions-refresh image analysis to test cronworkflow

fridex:


pacospace (commented):
Is it causing issues?

fridex:
> Is it causing issues? Should we have a way to enable disable them with env variables?

No, it looks like an RBAC issue probably when workflow-controller is not able to obtain the referenced templates. Closing this, it was used as an example reference.

/close

PR #2319, opened 2022-02-04
Increase document-sync deadline to 12 hours

fridex:
## Description

We have a lot of documents to sync, so let's increase the sync deadline and see how much time actually takes.


harshad16 (approved):
lgtm
thanks :100:

PR #2327, opened 2022-02-07
Increase memory for document-sync-job

fridex:
## Description

It looks like more memory is needed to sync some of the docs.


pacospace (approved):
lgtm! 💯

PR #2330, opened 2022-02-07
Provide port configuration to all deployment configs

fridex:

See https://github.com/thoth-station/user-api/issues/1645#issuecomment-1031857876

## Description

It looks like the base image no longer uses 8080 by default. Let's supply `PORT` environment variable to all deployment configs we have to make sure all web servers use 8080 without any regressions.


harshad16 (approved):
lgtm
thanks :100:

PR #2335, opened 2022-02-08
feat(prow): Enable override in open-services-group

tumido:
Override doesn't work in open-services-group. This should fix it.

https://github.com/open-services-group/community/pull/49#issuecomment-1032983342


harshad16 (approved):
lgtm
thanks :100:

PR #2345, opened 2022-02-14
Add missing dependency indexes env var for solver

fridex:


fridex:
Merging to make sure the functionality is in place.

PR #2356, opened 2022-02-15
Add a view for backlog refining to triage-party

codificat:

No

## Description

This adds rules to triage-party to show a list of issues that have `priority/backlog` or `priority/awaiting-more-evidence` so that they can be periodically reviewed.

Also restrict the `needs-triage` list to issues without a `triage/foo` label. This should not happen, but unfortunately it still does, and it clutters the view. The conflicting ones can be seen in the "inconsistencies" tab


codificat:
This is tested locally and works, so I am going to
/approve
to have it available online

Issue #2294, opened 2022-01-31
Merge Grafana dashboards of different deployments using variables


pacospace:
**Is your feature request related to a problem? Please describe.**
In order to avoid maintaining two or more json, let's merge dashboards using variables

**Describe the solution you'd like**
- identify variables across clusters
- merge dashboards
- add a description on top of each of the variables.

**Describe alternatives you've considered**
Have one dashboard type per deployment

**Additional context**

PR #2362, opened 2022-02-17
Merge overlays dashboards using variables

pacospace:

Related-To: https://github.com/thoth-station/thoth-application/issues/2294

This PR:
- Merges all dashboards with variables, so that no overlays are required and we can maintain one dashboard per class. (this is possible because Thanos default data source collect data from all overlays)
- Fixes labels in some dashbords (User-API dashboard -> `endpoint` -> `exported_endpoint`)
- Add README with description of all dashboards, how to add new ones and how to updated existing ones.


harshad16:
@pacospace have you got a chance to test this in grafana ?
(i know you would have created the dashboard in the grafana), asking from the perspective of the metrics, does the dashboard shows the metrics on the toggle between options.

pacospace:
> @pacospace have you got a chance to test this in grafana ? (i know you would have created the dashboard in the grafana), asking from the perspective of the metrics, does the dashboard shows the metrics on the toggle between options.

Yes @harshad16, I basically tested all dashboards and saved and json here. So once this is merged and the one in https://github.com/operate-first/apps/pull/1639 merged, we should see them up :) I can add the json again for showing how it looks like tomorrow at the call, wdyt?

harshad16:
awesome thanks for explaining.
I will merge it and maybe you can demo it for future reference as it is recorded :+1:

harshad16 (changes requested):
lgtm
thank again for the amazing pr.
if you could run the `pre-commit run` once , we could move forward with the pr merge.

codificat:
> we could move forward with the pr merge.

translating this to:
/approve

PR #2375, opened 2022-02-23
Increase adviser memory limit and requests

fridex:

Related: https://github.com/thoth-station/adviser/issues/2272

## Description

Let's increase memory requests and limits. I'm not sure if this will help, but worth trying. I think we will need to do some optimization here (either adjusting heuristics or checking memory consumption of objects on adviser side). This is a very first attempt to check what's wrong.

### DIFF
--- a/adviser/base/argo-workflows/advise.yaml
+++ b/adviser/base/argo-workflows/advise.yaml
@@ -235,10 +235,10 @@ spec:
         resources:
           limits:
             cpu: 1.1
-            memory: 6Gi
+            memory: 10Gi
           requests:
             cpu: 1.1
-            memory: 6Gi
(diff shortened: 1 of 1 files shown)


harshad16 (changes requested):
Ack on the change.
However this would be overrided as we patch resources in overlays 
please update this files : https://github.com/thoth-station/thoth-application/blob/master/adviser/overlays/aws-prod/advise-resources-patch.yaml

Also our test cluster doesn't have such resource, that why we have individual patch in the system.

fridex:
Thanks. I will try to experiment with one memory optimization in adviser. Let's see if it will work 👍🏻 

/close

Issue #2393, opened 2022-02-28
sprint production release v2022.03.14


harshad16:
Hello, Thoth-station!

This Issue would be used for the current sprint cycle production release.
By the end of the sprint cycle, we will consolidate the information of thoth-station components features upgrade and fixes in this issue.

/kind documentation
/area release-eng
/triage accepted
/milestone 2022.03.14
/priority important-soon

fridex:
**Notify users when they use PyTorch index**

Newly, the resolver notifies users if they consume releases from the PyTorch index.

* https://github.com/thoth-station/adviser/pull/2279

fridex:
**Memory optimizer in adviser**

If Thoth's adviser consumes too much memory and is killed on OOM in a deployment, a memory optimizer can be turned on. The memory optimizer removes data from some of the internal data structures to optimize memory consumption. This can slow down to finding a resolved set of dependencies.

* https://github.com/thoth-station/adviser/issues/2203
* https://github.com/thoth-station/adviser/pull/2280

harshad16:
we have completed the release for v2022.02.28 :tada: :confetti_ball: :partying_face: 

### Features

**Notify users when they use PyTorch index**

Newly, the resolver notifies users if they consume releases from the PyTorch index.

* https://github.com/thoth-station/adviser/pull/2279

**Memory optimizer in adviser**

If Thoth's adviser consumes too much memory and is killed on OOM in a deployment, a memory optimizer can be turned on. The memory optimizer removes data from some of the internal data structures to optimize memory consumption. This can slow down finding a resolved set of dependencies or making the search not that optimal. Users, however, can get results instead of OOM.

* https://github.com/thoth-station/adviser/issues/2203
* https://github.com/thoth-station/adviser/pull/2280
* https://github.com/thoth-station/thoth-application/pull/2395

**Created new handler that uses image analysis results and generate new prescriptions**

Automatically propagated from the container image analyses - the container image analysis would provide how Pipfile looked like on image build and that information will be part of the image analyses information derived by package-extract (similarly as we propagate information about RPM packages). In this case, we implemented a handler in prescriptions-refresh-job that:
* checks what ps images are hosted on quay
* ask what is the container image analysis result and how the Pipfile looked like during the build
* automatically create prescriptions based on direct dependencies in Pipfile

References:
* https://github.com/thoth-station/core/issues/343
* https://github.com/thoth-station/prescriptions-refresh-job/issues/87
[...]

PR #2394, opened 2022-02-28
Reduce CPU used by adviser on AWS prod

fridex:
## Description

Reduce CPU allocated per adviser run. Adviser uses 2 processes, one is CPU expensive, the other one is not that CPU expensive. 2.5 CPU look too much to allocate per an adviser run.

### DIFF
--- a/adviser/overlays/aws-prod/advise-resources-patch.yaml
+++ b/adviser/overlays/aws-prod/advise-resources-patch.yaml
@@ -3,7 +3,7 @@
   value:
     requests:
       memory: "6Gi"
-      cpu: "2.5"
+      cpu: "1.1"
     limits:
       memory: "6Gi"
-      cpu: "2.5"
+      cpu: "1.1"


fridex:
/assign @harshad16 

Do you have some metrics on this? I think 2.5 CPU is too much to allocate in this case.

harshad16:
> Do you have some metrics on this? I think 2.5 CPU is too much to allocate in this case.

we don't have direct metrics dashboard in grafana
however if when user executes an advisor run, we can see the CPU metrics like this:
https://console-openshift-console.apps.balrog.aws.operate-first.cloud/monitoring/query-browser?query0=sum%28kube_pod_resource_limit%7Bresource%3D%27cpu%27%2Cpod%3D%27adviser-220303140706-cdd2fa9faf996ade-1132623225%27%2Cnamespace%3D%27thoth-backend-prod%27%7D%29&query1=sum%28kube_pod_resource_request%7Bresource%3D%27cpu%27%2Cpod%3D%27adviser-220303140706-cdd2fa9faf996ade-1132623225%27%2Cnamespace%3D%27thoth-backend-prod%27%7D%29+BY+%28pod%2C+namespace%29
it seems like it utilize or at least touch 2.5m CPU

harshad16:
Interesting find ,cpu is not being used much 
however memory is going to oom for stack like this: https://github.com/aicoe-aiops/ocp-ci-analysis/blob/master/Pipfile
adviser-run

goern:
@fridex just out of curiosity, can we correlate the incr in memory usage with the source code allocating the memory?

fridex:
@goern Memory management should be improved with the recent changes - https://github.com/thoth-station/thoth-application/issues/2393#issuecomment-1054582150 The changes should be part of the next release. We were not observing any OOM in the stage environment with the linked change.

@harshad16 I would expect CPU to be used in later phases of the resolution process when the adviser does not talk to the database that much and rather uses cached results. Given the implementation, I would be very surprised if the implementation would use more than 1.5 CPU - we spawn two processes, one just makes sure proper memory management is done and controls child (ex. when the CPU time exceeds).

harshad16 (approved):
lgtm
thanks :100:

goern:
> @fridex just out of curiosity, can we correlate the incr in memory usage with the source code allocating the memory?

I was more wondering if we are able to link both: metrics and source code.

PR #2398, opened 2022-02-28
Increase memory for prescriptions-refresh-pg

fridex:

The current deployment fails on OOM.


harshad16 (approved):
lgtm
thanks :100:

PR #2401, opened 2022-03-01
Increase deadline for prescriptions-refresh jobs

fridex:

Some jobs were killed as they exceeded deadline.


mayaCostantini (approved):
lgtm 💯

PR #2402, opened 2022-03-01
Increase memory for prescriptions-refresh jobs to 1Gi to avoid OOM

fridex:

Avoid OOM. We can further reduce memory allocation based on metrics.


mayaCostantini (approved):
lgtm 💯

PR #2404, opened 2022-03-01
Run all the steps in prescriptions-refresh PyPI workflow

fridex:
## Description

We have run only the very first template.


mayaCostantini (approved):
lgtm 👍🏻 💯

PR #2408, opened 2022-03-01
Adjust adviser's memory optimizer limit

fridex:
## Description

It looks like we have still some memory available to fill. Activate the memory optimization later.

image

image

### DIFF
--- a/adviser/base/argo-workflows/advise.yaml
+++ b/adviser/base/argo-workflows/advise.yaml
@@ -123,7 +123,7 @@ spec:
           - name: THOTH_SENTRY_IGNORE_LOGGER
             value: "thoth.adviser.run"
           - name: THOTH_ADVISER_MEM_OPTIMIZER_LIMIT
-            value: "5000000"
+            value: "5750000"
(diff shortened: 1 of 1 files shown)


fridex:
For future selves: adviser can suddenly do a memory consumption bump when it is aggregating results as some pipeline units are called after the actual resolution process (wraps). That's why there is left some space for the process to aggregate resolutions for users.

mayaCostantini:
> For future selves: adviser can suddenly do a memory consumption bump when it is aggregating results as some pipeline units are called after the actual resolution process (wraps). That's why there is left some space for the process to aggregate resolutions for users.

@fridex Should we document this somewhere?

mayaCostantini (approved):
lgtm 💯 👍🏻

fridex:
> > For future selves: adviser can suddenly do a memory consumption bump when it is aggregating results as some pipeline units are called after the actual resolution process (wraps). That's why there is left some space for the process to aggregate resolutions for users.
> 
> @fridex Should we document this somewhere?

Good idea, opened https://github.com/thoth-station/adviser/pull/2289 PTAL 👍🏻

Issue #2367, opened 2022-02-17
Enable CI on OSC cluster 


MichaelClifford:
We need CI to execute, check pre-commit and do merge actions in our repositories in the OS-Climate organization, for example: https://github.com/os-climate/aicoe-osc-demo  

@harshad16

goern:
This would require Prow receiving webhooks from the repos or the org, and being able to act on the PRs. So please configure a webhook like in the screenshot and invite https://github.com/sesheta as an owner to the org. Please ping this issue if you have invited Sesheta, as we need to accept the invite.

![image](https://user-images.githubusercontent.com/260331/155086661-5d6b0dc5-2a28-441f-b293-ec40127ec1d9.png)

/priority critical-urgent
/assign @harshad16 
/triage accepted

goern:
any update on this?

MichaelClifford:
@goern @harshad16 What secret do I use for this webhook?

goern:
it should be https://github.com/thoth-station/thoth-application/blob/master/prow/overlays/smaug/secrets.enc.yaml#L23-L29
@harshad16 ?

harshad16:
will pass the secret to @MichaelClifford personally :+1:

MichaelClifford:
closed by #2412 and #2413

PR #2412, opened 2022-03-04
add os-climate base config/plugins

goern:
/Cc @MichaelClifford


goern:
related-to: #2367

PR #2418, opened 2022-03-08
Do not send message by default after Amun inspections

fridex:
## Description

Do not send messages after Amun inspections. We have this part of integration off.


mayaCostantini (approved):
lgtm 👍🏻

PR #2421, opened 2022-03-08
Revert "Do not send message by default after Amun inspections"

fridex:
Reverts thoth-station/thoth-application#2418

The default behavior is encoded in sources:

https://github.com/thoth-station/common/blob/5c7452c5140a0105acfa6260ffa244d84c1f7431/thoth/common/openshift.py#L867-L869

We should make this configurable in templates https://github.com/thoth-station/common/pull/1231

As this change has no effect in the deployment, reverting. Let's fix sending messages step.


harshad16 (approved):
lgtm
thanks :100:

PR #2436, opened 2022-03-14
Run graph-refresh every 40 mins, instead of every hour

harshad16:
Run graph-refresh every 40 mins, instead of every hour

gr
oE/aicoe-cd` and Prod `operate-first/argocd-apps`.

## Description

The workload is getting resolved by every 40 mins, and the metrics are alerting we have compute power unused.
so making the graph-refresh job run for every 40 mins in stage cluster.

### DIFF
--- a/graph-refresh/base/cronjob.yaml
+++ b/graph-refresh/base/cronjob.yaml
@@ -7,7 +7,7 @@ metadata:
     app: thoth
     component: graph-refresh
 spec:
-  schedule: "0 */1 * * *"
+  schedule: "*/40 * * * *"
   suspend: true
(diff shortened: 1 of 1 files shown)


harshad16:
cc: @fridex 
please review this.
This would only effect stage and test(suspended for regular run).

fridex (approved):
looks good to me, thanks! 💯

fridex:
It might be that we do not have any packages to be solved (metrics suggest it) - hence the alarm. Do we want to plug Fedora 35 solver? CC @Gregory-Pereira

harshad16:
ack, maybe in this week , we can plug it in  :+1:

PR #2454, opened 2022-03-21
Propagate URL to Thoth Search UI in deployments

mayaCostantini:

Related to https://github.com/thoth-station/user-api/issues/1721

### DIFF
--- a/user-api/overlays/moc-prod/deploymentconfig.yaml
+++ b/user-api/overlays/moc-prod/deploymentconfig.yaml
@@ -128,7 +128,7 @@ spec:
                   key: secret-key
             - name: THOTH_SEARCH_UI_URL
               valueFrom:
-                secretKeyRef:
+                configMapKeyRef:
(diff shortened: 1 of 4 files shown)


user-api/overlays/moc-prod/deploymentconfig.yaml
```diff
@@ -126,6 +126,11 @@ spec:
                 secretKeyRef:
                   name: ceph
                   key: secret-key
+            - name: THOTH_SEARCH_UI_URL
+              valueFrom:
+                secretKeyRef:
+                  name: thoth-search-ui-url
+                  key: "https://thoth-station.ninja/search/"
```
fridex, line 133: Is this the key we want to use for the secret?

mayaCostantini: Right thank you (copy-paste error), I will replace it with `configMapKeyRef` 👍🏻

PR #2456, opened 2022-03-21
Fix user-api Thoth Search UI environment variable in deployments

mayaCostantini:
/assign @harshad16


harshad16 (approved):
lgtm
thanks for fixing it :100:

PR #2457, opened 2022-03-22
update all prow imagestreamtags to latest builds

goern:


prow/overlays/smaug/imagestreamtags.yaml
```diff
@@ -57,6 +57,11 @@ spec:
       kind: DockerImage
       name: gcr.io/k8s-prow/clonerefs:v20211027-fe152eb205
```
Gregory-Pereira, line 58: Any reason that we keep the old ImageStream tags?

PR #2459, opened 2022-03-23
Define a restartPolicy for prescriptions-refresh-gh containers

mayaCostantini:

Related to https://github.com/thoth-station/prescriptions-refresh-job/issues/130

## Description

Define a `restartPolicy` if containers for prescriptions-refresh-job using a GitHub token fail when the token limit is reached.


mayaCostantini:
@harshad16 PTAL

harshad16 (changes requested):
seems like the workflow template container doesnt have these field configurable.
https://argoproj.github.io/argo-workflows/fields/#container

please close the pr.
The default restart policy is set to `onfailure` , so we are good on that front
what we have to do is make the app fail with a proper exit condition 
so this is not needed, change might be needed on the prescription front.

PR #2463, opened 2022-03-28
set mi-scheduler-cm configmap for mi-scheduler cronjob

harshad16:
set mi-scheduler-cm configmap for mi-scheduler cronjob

## Description

we need this separation as the config map mi-scheduler lives in another namespace as thoth-common reads it and is used in workflows, so we need this as separate as it is directly being used in cronjob. 
https://github.com/thoth-station/thoth-application/blob/426358d5f6fad745e35a99ab74ee6cbf7067b2b6/mi-scheduler/base/cronjob.yaml#L63
https://github.com/thoth-station/thoth-application/blob/426358d5f6fad745e35a99ab74ee6cbf7067b2b6/mi-scheduler/overlays/moc-prod/kustomization.yaml#L24


harshad16:
/approve
Merging it

PR #2468, opened 2022-03-28
configmap amun is required by both amun-api and amun-wf

harshad16:
configmap amun is required by both amun-api and amun-wf

## Description

- The configmap amun is used by both amun-api server and graph-sync workflow
- workflowtemplate requires to present in thoth-amun-inspection namespace


harshad16:
/approve
Merging it

PR #2504, opened 2022-04-07
Bump integration-tests to v0.11.1

fridex:

This should fix `count` and `limit` in queries.


goern:
oh thanks, have overseen that!
:eye: :100: 
/approve

Issue #2513, opened 2022-04-12
Managing data across deployments


fridex:
**Is your feature request related to a problem? Please describe.**

As Thoth operator, I would like to make sure data are properly managed across deployments. As of now, we use a staging environment to compute data (given the resources available) and propagate a database dump to prod environments. This situation proved to be not scalable long-term as production can write into the database as well and we can get out of sync easily. If we overwrite the prod environment with staging data, we can also lose some information.

One of the proposed solutions discussed was to do updates per-table. This solution can introduce overhead and possible inconsistencies we should avoid (ex. package entries in the database created by solvers can be overwritten by packages detected using container image analyses).

---

Another solution is to let syncs happen on the programming level. In other words, we could keep our background jobs that copy data from staging environment to production running (document-sync-job). In that case, the job places documents on ceph in the production environment. A subsequent graph-sync job can sync these data into the database so that they are available in prod (even if they were computed in the staging environment). This approach seems to be scalable and might require less maintenance.

- [ ] make sure the sync method in thoth-storages can randomize document listing so that we can run multiple sync jobs in parallel
- [ ] make sure the sync-job uses the adjusted method and we can turn the job into a cronjob that periodically syncs data into the database (multiple sync jobs can be part of the cronjob to support parallel syncs)
- [ ] make sure the sync job can be also triggered from management-api
- [ ] make sure a user can select what documents should be synced on management api (ex. data for fedora 35 running python 3.10)
- [ ] make sure database dumps are created also in the production environment

fridex:
CC @Gregory-Pereira @harshad16

harshad16:
Thank you @fridex for the issue with all details

Gregory-Pereira:
>     * [ ]  make sure the sync-job uses the adjusted method and we can turn the job into a cronworkflow that periodically syncs data into the database (multiple sync jobs can be part of the cronworkflow to support parallel syncs)
I think Maya's current PR should update everything in such a way in `storages` that the only changes needed in the `sync-job` are to bump the version for storages once it goes through, and construct a `CronWorkflow` from the `CronJob template` in [openshift.yaml](https://github.com/thoth-station/sync-job/blob/master/openshift.yaml).

PR #2524, opened 2022-04-20
creating cron-workflow to invoke job template for the sync-job

Gregory-Pereira:

Partially Addresses: https://github.com/thoth-station/thoth-application/issues/2513
Related to: https://github.com/thoth-station/sync-job/pull/61

## Description

Creating a cron-workflow for the sync-job

/cc @harshad16

### DIFF
--- a/sync-job/base/cron-workflow.yaml
+++ b/sync-job/base/cron-workflow.yaml
@@ -1,0 +1,45 @@
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: sync-job-cw
+  labels:
+    app: thoth
+    component: sync
+spec:
+  schedule: "0 10 14 * *"
+  timezone: "Etc/UTC"
+  concurrencyPolicy: "Replace"
+  startingDeadlineSeconds: 0
+  workflowSpec:
+    entrypoint: DocumentSyncTypeOne
+    templates:
+    - name: DocumentSyncTypeOne
+      steps:
+      - - name: SyncDocuments
+          templateRef:
+            name: sync-job-wt
+            template: SyncDocumentsByType
+          arguments:
+            parameters:
+              - name: THOTH_DOCUMENT_TYPE
+                value: "solver"
+    - name: DocumentSyncTypeTwo
+      steps:
+      - - name: SyncDocuments
+          templateRef:
+            name: sync-job-wt
+            template: SyncDocumentsByType
+          arguments:
+            parameters:
+              - name: THOTH_DOCUMENT_TYPE
+                value: "revsolver"
+    - name: DocumentSyncTypeThree
+      steps:
+      - - name: SyncDocuments
+          templateRef:
+            name: sync-job-wt
+            template: SyncDocumentsByType
+          arguments:
+            parameters:
+              - name: THOTH_DOCUMENT_TYPE
+                value: "security-indicator"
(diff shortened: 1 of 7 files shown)


Gregory-Pereira:
I know that when going from stage to production we should be syncing adviser, revsolver, and security-indicator documents, but what about from production to stage?

sync-job/base/cron-worfklow.yaml
```diff
@@ -0,0 +1,26 @@
...
+  concurrencyPolicy: "Replace"
+  startingDeadlineSeconds: 0
+  workflowSpec:
+    paralellism: 4
+    entrypoint: sync-job-wt
+    templates:
+      container:
+        image: quay.io/thoth-station/sync-job:v0.2.3-dev
+        command: ["oc", "process"]
+        args: ["-p IMAGE_VERSION=$IMAGE_VERSION", "-p THOTH_SYNC_FORCE=$THOTH_SYNC_FORCE", "-p THOTH_SYNC_GRACEFUL=$THOTH_SYNC_GRACEFUL", "-p THOTH_SYNC_DEBUG=$THOTH_SYNC_DEBUG", "-f openshift.yaml", " | oc apply -f -"]
```
harshad16, line 17: This is not needed.
There is concept of source-to-image , the entrypoint is already set to the run script that can `app.py` and others based on the programming language. The pointer to different script can be done via setting env vars
https://docs.openshift.com/dedicated/3/using_images/s2i_images/python.html#using-images-python-configuration

sync-job/base/cron-worfklow.yaml
```diff
@@ -0,0 +1,26 @@
...
+        env:
+          - name: IMAGE_VERSION
+            value: v0.2.3-dev
```
harshad16, line 20: This is not needed.
this was used on the sync-job template and is not a part of the application itself. 
https://github.com/thoth-station/sync-job/blob/f630e32b9b62b0e81879a16f95ea34b7e1dadce8/openshift.yaml#L79

sync-job/base/openshift.yaml
```diff
@@ -0,0 +1,161 @@
...
+                    secretKeyRef:
+                      name: postgresql
+                      key: database-name
+              resources:
+                limits:
+                  memory: "2Gi"
+                  cpu: 4
+                requests:
+                  memory: "2Gi"
+                  cpu: 4
```
harshad16, line 161: The workflow would do the work of the job, 
we won't require a separate job.
the cronworkflow would schedule workflow as needed. 

Move all these env vars to the cronworkflow.

harshad16 (changes requested):
good start
i have added some suggestions.

Gregory-Pereira:
It uses 3 templates because both for prod and stage, there are 3 separate types of documents to be synced. Additionally I can apply the resources (it is correct YAML syntax), however, the workflow controller rejects this `CronWorkflow` by giving me the following error: `"Unable to convert unstructured to CronWorkflow when syncing CronWorkflows" error="cannot restore slice from map"`

Gregory-Pereira:
Cron-workflow is now accepted, ready for review.

Gregory-Pereira:
Attempting to implement this with a cron-workflow which uses a workflow template so the changes be used to both invoke a cronworkflow to run sync-job periodically, as well as be invoked from the management api via workflow. That being said I am encountering an Argo issue: `template reference exceeded max depth (10)`.

Gregory-Pereira:
Issue about template reference exceeding max depth has been resolved, hoping that @harshad16 could take a look when you have time.

harshad16 (approved):
/lgtm
/approve
thanks :100:

PR #2543, opened 2022-05-04
feat(prow): OSG Peribolos needs Tide stop skipping checks

tumido:
Tide is skipping contexts not defined by Prow config (enabled in settings). Turn this off for Peribolos as a service repo.


harshad16 (approved):
lgtm :+1:

Issue #2556, opened 2022-05-11
[1pt] Give user-api secret creation and editing permissions


KPostOffice:
**Is your feature request related to a problem? Please describe.**
Webhook implementation currently proposed requires new permissions. Creating a secret for adviser runs and the ability to patch these secrets.

**Describe the solution you'd like**
Add secret creation and editing to user-api role
https://github.com/thoth-station/thoth-application/blob/master/user-api/base/role.yaml

PR #2585, opened 2022-06-06
add create, and update permissions for secret resources 4 user-api

KPostOffice:
I gave user-api its own rule, removing it from the rule with
configmap permissions. I then duplicated the permissions that were
already present, and added "create" and "update" verbs to the rule.

closes: https://github.com/thoth-station/thoth-application/issues/2556


harshad16 (approved):
/lgtm

PR #2590, opened 2022-06-13
Suspend scorecards handler and schedule refresh job monthly

mayaCostantini:
## Description

- Temporarily disable the prescriptions-refresh-job handler responsible for updating scorecards as it consumes too many resources
- Schedule this cronjob to be run monthly instead of weekly


harshad16 (approved):
/lgtm
/approve

PR #2603, opened 2022-07-03
Temporarily suspend prescriptions

mayaCostantini:
## Description

Temporarily disable prescriptions because of an important increase in BigQuery API costs for some handlers.


mayaCostantini:
/approve
Merging as a fix

PR #2605, opened 2022-07-06
[prow] Add branch cleaner plugin configuration

codificat:

Fixes thoth-station/prescriptions-refresh-job#172

Just a deployment of the updated prow config.

## Description

This configures the `branchcleaner` [prow plugin](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/branchcleaner) so that source branches for merged PRs between two branches on the same repository are automatically deleted.

This can help with e.g. Kebechet PRs or prescription-refresh-job runs.

### DIFF
--- a/prow/overlays/smaug/plugins.yaml
+++ b/prow/overlays/smaug/plugins.yaml
@@ -229,6 +229,30 @@ project_manager:
               org: thoth-station
               state: open
 
+branch_cleaner:
+  preserved_branches:
+    AICoE:
+      - main
+      - master
+    aicoe-aiops:
+      - main
+      - master
+    b4mad:
+      - main
+      - master
+    open-services-group:
+      - main
+      - master
+    operate-first:
+      - main
+      - master
+    os-climate:
+      - main
+      - master
+    thoth-station:
+      - main
+      - master
+
 plugins:
   b4mad:
     plugins:


harshad16:
/hold

before this can be merged, we have to update the branch-protection config, or else they both would fight for race condition.

prow/overlays/smaug/plugins.yaml
```diff
@@ -229,6 +229,30 @@ project_manager:
...
+      - master
+    open-services-group:
+      - main
+      - master
+    operate-first:
+      - main
+      - master
+    os-climate:
+      - main
+      - master
```
harshad16, line 251: we should start with only thoth-station repo, as we are using this for the first time.

codificat: if I understand the [configuration of the plugin](https://github.com/kubernetes/test-infra/blob/e831f6aee73552c33fdc07c5b5ba7dedb29497b4/prow/plugins/plugin-config-documented.yaml#L73-L84) correctly, there is no way to restrict its operation to a specific org... you can configure *branch names* to be preserved org-wide (which is what I did there for `main` and `master`, for all repos on all orgs), but my understanding is that if something is not listed here, then it is not preserved.

harshad16:
@codificat please move this pr to https://github.com/operate-first/apps/tree/master/prow
and close this one 
we have migrated prow operations to operate-first.

harshad16:
closing this pr, as it needs to be made in operate-first/apps repo.
Thank you.

Issue #2557, opened 2022-05-11
[2pt] Create a new ServiceAccount for callback sending task with the ability to delete secrets


KPostOffice:
**Is your feature request related to a problem? Please describe.**
After callback urls are sent, the created secret needs to be removed.

**Describe the solution you'd like**
Create a new serviceAccount, role, roleBinding for a new workflow-helpers task which sends results to callback url

PR #2607, opened 2022-07-06
add new role and role binding for adviser

KPostOffice:
add role which will be used for adviser workflows so that secrets created for callback urls can be deleted

closes: #2557 

### DIFF
--- a/core/base/kustomization.yaml
+++ b/core/base/kustomization.yaml
@@ -9,3 +9,4 @@ resources:
   - imagestreams.yaml
   - argo-workflows/send-message.yaml
   - argo-workflows/send-messages.yaml
+  - send-webhooks-sa.yaml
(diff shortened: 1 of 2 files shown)


adviser/base/adviser-sa.yaml
```diff
@@ -0,0 +1,29 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: adviser
```
harshad16, line 5: which component would use this sa ?

KPostOffice: this will be in the adviser workflow. I was looking into it though and I think I can patch the pod definition for a single workflow step so that we don't have to run the whole adviser workflow with elevated priviledges

harshad16 (approved):
/lgtm
/approve

PR #2631, opened 2022-08-23
Fix cache expiration value

mayaCostantini:
It had been set to 12 hours in the configmap. 
The default value when this environment variable is not present for user api is 3 hours, but we can leave it as 4 hours here if needed.


harshad16 (approved):
/lgtm
/approve

PR #2641, opened 2022-08-29
Add a rule to triage-party to highlight issues without a SIG

codificat:
This is to add a rule in the weekly triage view of triage-party to show any issues that do not yet have a SIG label.

### DIFF
--- a/triage-party/overlays/moc/config.yaml
+++ b/triage-party/overlays/moc/config.yaml
@@ -102,6 +102,7 @@ collections:
     rules:
       # Issues pending triage
       - issue-needs-triage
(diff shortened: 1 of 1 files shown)


codificat:
Also taking the opportunity to add CNBi-related repositories for triage-party to watch for.

Unfortunately, though, these repos (and the other AICoE already listed) do not have `sig/*` labels available... so, issues in these repos will, for the time being, always show up in the *needs SIG* view.

codificat:
I am going to go ahead and self-
/approve
this one to see if the new view helps with our upcoming backlog refinement session.

Issue #2642, opened 2022-08-30
Ingestion workflow having issue submitting response to ceph datastore


harshad16:
**Describe the bug**
The data ingestion is carried out with the solver runs, which are created with agro-workflows.
workflow-controller schedules the argo workflows based on the description, as workflows are constructed with different steps, each step is scheduled as a pod based on the description.

first workflows which get schedules, stores the result in ceph store and next step execute based on the results.
However the workflow is failing to store the information in ceph store.
```
time="2022-08-30T17:27:24.419Z" level=info msg="Saving from /mainctrfs/mnt/workdir/revsolver-220830170029-77e3b349bee4c259 to s3 (endpoint: s3.upshift.redhat.com, bucket: thoth, key: data/ocp4-stage/revsolver/revsolver-220830170029-77e3b349bee4c259)"
time="2022-08-30T17:27:24.437Z" level=warning msg="Failed to put file: Get http://s3.upshift.redhat.com/thoth/?location=: dial tcp 10.0.239.1:80: connect: connection refused"
```
the workflow seems to be sending out info to http insecure instead of secure endpoint.

**To Reproduce**
1. Go to `thoth-middletier-stage`
2. Click on solver workflows
3. See error in the wait container.

**Expected behavior**

successful execution of the solver workflows.

harshad16:
The ceph storage in stagging cluster, seems to have deprecated the http insecure for s3.upshift ceph.
we have to re-adjust to use the secure https for connection.

harshad16:
Fixed the issue:
![Screenshot from 2022-08-30 15-19-17](https://user-images.githubusercontent.com/14028058/187534794-5a390d8c-7b7b-40a1-aced-f46a2cad37b6.png)

PR #2643, opened 2022-08-30
:fire: switch insecure false for the agro-worflows

harshad16:
switch insecure false for the agro-worflows

- Removed insecure flag from worklfows

Fixes: #2642 

## Description

- The argo-workflow artifact overrides the workflow-controller configuration. 
so we have removed the insecure bit from workflow and are going to rely on the workflow controller configuration.


harshad16:
Merging this to fix the ingestion issue.
/approve

PR #2661, opened 2022-09-26
Use emptydir for db pg_dump

harshad16:
Use emptydir for db pg_dump

Related-to: https://github.com/thoth-station/graph-backup-job/issues/242


VannTen (changes requested):
Can we use emptyDir on all the environments or some don't have enough local storage ?
If we can, I would move the volume def in base/.

Besides that, I think we should add a `resources.requests.ephemeral-storage` matching the expected size of the db dumb. (If it's available. It's a beta k8s feature, not sure if exposed on the openshift versions we have)

graph-backup-job/overlays/aws-prod/kustomization.yaml
```diff
@@ -23,26 +22,8 @@ patches:
...
-      group: batch
-      version: v1beta1
-      kind: CronJob
-      name: graph-backup
-  - patch: |-
-      - op: add
-        path: "/spec/jobTemplate/spec/template/spec/volumes/-1"
-        value: {"name": "postgresql-data","persistentVolumeClaim":{"claimName": "dbbackup"}}
+        path: "/spec/jobTemplate/spec/template/spec/containers/0/resources"
+        value: {"requests": {"memory":"1Gi", "cpu": "1", "ephemeral-storage": "20Gi"}, "limits": {"memory":"1Gi", "cpu": "1", "ephemeral-storage": "20Gi"}}
```
VannTen, line 26:
```suggestion
        path: "/spec/jobTemplate/spec/template/spec/containers/0/resources/requests/ephemeral-storage"
        value: "20Gi"
```

VannTen:
```suggestion
      - op: add
        path: "/spec/jobTemplate/spec/template/spec/containers/0/resources/limits/ephemeral-storage"
        value: "20Gi"
```

VannTen: I would only add the ephemeral in the patch (same for other environment).
Or maybe we can add 20Gi as the base and only override test at 1Gi ?

Side note: Do we want to add the limit on ephemeral-storage ? The idea is to be alerted immediately if db dump starts to go over 20Gi ?

VannTen: (not sure if my second sugg applies cleanly on top on the other, but the idea : two add ops specifying the value only at the correct path.

VannTen (changes requested):

graph-backup-job/overlays/aws-prod/kustomization.yaml
harshad16: Ack, i have made the changes as you suggested.
The limits is something requested by cluster team so fit with the namespace limits.

about we being alerted , the exceed of those limit would cause failure in system, so we should be notified.

graph-backup-job/overlays/moc-prod/kustomization.yaml
```diff
@@ -20,3 +20,15 @@ patches:
       version: v1
       kind: Job
       annotationSelector: "operation=chat-notification"
+  - patch: |-
+      - op: add
```
VannTen, line 24:
```suggestion
      - op: replace
```

graph-backup-job/overlays/moc-prod/kustomization.yaml
```diff
@@ -20,3 +20,15 @@ patches:
...
+        path: "/spec/jobTemplate/spec/template/spec/containers/0/resources/requests/ephemeral-storage"
+        value: "20Gi"
+      - op: add
```
VannTen, line 27:
```suggestion
      - op: replace
```

graph-backup-job/overlays/ocp4-stage/kustomization.yaml
```diff
@@ -23,26 +22,11 @@ patches:
       annotationSelector: "operation=chat-notification"
   - patch: |-
       - op: add
```
VannTen, line 24:
```suggestion
      - op: replace
```

graph-backup-job/overlays/ocp4-stage/kustomization.yaml
```diff
@@ -23,26 +22,11 @@ patches:
...
-        value: {"name": "postgresql-data","mountPath": "/var/lib/pgsql"}
-    target:
-      group: batch
-      version: v1beta1
-      kind: CronJob
-      name: graph-backup
-  - patch: |-
+        path: "/spec/jobTemplate/spec/template/spec/containers/0/resources/requests/ephemeral-storage"
+        value: "20Gi"
       - op: add
```
VannTen, line 27:
```suggestion
      - op: replace
```

VannTen (changes requested):
I've gone read the json patch docs and I think we need `replace` rather than `add` since we have a base value now ?
For the rest it's
/lgtm

graph-backup-job/overlays/ocp4-stage/kustomization.yaml
harshad16: Both works,  anyhow i have changed it :+1:

PR #2722, opened 2023-01-07
Suspend prescription refresh execution

codificat:

Currently open PRs in the prescriptions repository

## Description

This is to temporarily stop the prescription refresh jobs.

While looking at the last execution of the job and the generated PRs I found out that current prescriptions seem to fail validation.

Pausing the job (meant to execute later today) until we clarify this.


codificat:
It's the weekend and I feel we should pause this before it runs later today, so... going to merge manually :pray:

PR #2741, opened 2023-02-10
:sleeping: Turn off sentry logging by removing sentry_dsn

harshad16:
:sleeping: Turn off sentry logging by removing sentry_dsn

## Description

As part of the hibernation of the project.

### DIFF
--- a/amun/overlays/moc-prod/amun-api/secrets.enc.yaml
+++ b/amun/overlays/moc-prod/amun-api/secrets.enc.yaml
@@ -5,16 +5,16 @@ items:
       metadata:
         name: amun
       data:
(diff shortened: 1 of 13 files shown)


goern:
as `sentry_dsn` is removed from the secrets, it will not be injected as an env var in deployments, and therefore thoth.logging will not send to sentry. correct?

harshad16:
> as `sentry_dsn` is removed from the secrets, it will not be injected as an env var in deployments, and therefore thoth.logging will not send to sentry. correct?

as the value of `sentry_dsn` is removed, the env var SENTRY_DSN would fail in filter condition here
https://github.com/thoth-station/common/blob/48e0bea82abccc4de312b193dd8771af7b8d4bf1/thoth/common/logging.py#L301
and would be noted as turned off https://github.com/thoth-station/common/blob/48e0bea82abccc4de312b193dd8771af7b8d4bf1/thoth/common/logging.py#L340

PR #2742, opened 2023-02-10
:sleeping: Slow down the daily jobs to monthly on stage cluster

harshad16:
:sleeping: Slow down the daily jobs to monthly on-stage cluster

## Description

- slow down the daily jobs to monthly 
- push back graph-refresh-job to hourly

### DIFF
--- a/graph-backup-job/base/cronjob.yaml
+++ b/graph-backup-job/base/cronjob.yaml
@@ -7,7 +7,7 @@ metadata:
     app: thoth
     component: graph-backup
 spec:
-  schedule: "@daily"
+  schedule: "@monthly"
   suspend: true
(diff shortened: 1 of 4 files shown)


goern:
@harshad16 why do we want to do this? is it generating a lot of maintenance work/errors/... ?

harshad16:
> @harshad16 why do we want to do this? is it generating a lot of maintenance work/errors/... ?

As we are not planning to do active development 
the system like integration-tests and metrics components which provide inclusion data, would not be looked upon
based on this, I moved it from daily to monthly.

the graph-refresh is moved from 40mins to 60mins, as the ingestion rate-of-generation seems to high than rate-of-consumption.

PR #2746, opened 2023-02-16
remove deployment of Thoth's traige party

goern:


goern:
see also https://github.com/operate-first/apps/pull/2737

codificat:
> see also [operate-first/apps#2737](https://github.com/operate-first/apps/pull/2737)

Isn't that enough? (i.e. remove the deployment itself)

This PR removes all configuration, which I believe can stay around in case it becomes useful again. The app can be re-deployed in the future again based on it, no?

harshad16 (approved):
/lgtm

As the GitHub history is persisted, in the future, the commit can be reverted to bring back the triage party.

PR #2752, opened 2023-07-31
Updated the kafka certs for with kafka change

harshad16:
Updated the kafka certs for with kafka change
The kafka deployment was failing causing , the whole system be down.
Screenshot from 2023-07-30 21-35-46


harshad16:
Merging to get the system fixed.

/approve.

